PR #55, opened 2021-04-30
added extra parameter to the SeQuant context which indentifies whether…

cmasteran1:
… or not spin free or spin basis is used for fermions. Context affects the symmetry of operators and the printing labels to_latex()


evaleev:
@cmasteran1 just to make sure: this depends on #58 ?

cmasteran1:
Yes those changes only make sense if the wick algorithm uses the tensor form

examples/uccf12/uccf12.cpp
```diff
@@ -78,38 +79,89 @@ try_main() {
 
   // single commutator, needs symmetrization
   {
-    auto h = H();
+    auto h = H(false);
     auto r = R12(gg_space);
-    auto hr_comm = do_wick( ex<Constant>(2) * (h*r - r*h) );  // this assumes symmetrization includes 1/2
+    /*auto hr_comm = do_wick( ex<Constant>(2) * (h*r - r*h) );  // this assumes symmetrization includes 1/2
```
evaleev, line 84: is this meant to be removed? Commented out code should be removed, unless used as documentation of something

examples/uccf12/uccf12.cpp
```diff
@@ -47,7 +48,7 @@ try_main() {
     wick.spinfree(false)
         .full_contractions(false);
     auto result = wick.compute();
-    simplify(result);
+   simplify(result);
```
evaleev, line 51: formatting seems off here and elsewhere ... do you not have auto-formatting ON in IDE?

evaleev (changes requested):

PR #56, opened 2021-05-12
Open-shell spin-trace function

nakulteke:
The function generates a vector of expressions corresponding to spin blocks of the residual equation.
For eg, in the case of CCSD, R1->{⍺, β} blocks; R2->{⍺⍺, ⍺β, β⍺, ββ} blocks.

PR #57, opened 2021-05-12
Factorization

bimalgaudel:
should be merged before #56 to preserve spin.hpp decomposition into spin.hpp and spin.cpp to prevent ODR violations.


evaleev (changes requested):
use the established naming convention + dox

PR #60, opened 2021-06-26
begin and end changed to ranges::{begin,end}

bimalgaudel:


evaleev:
@bimalgaudel needs header incl

bimalgaudel:
> @bimalgaudel needs header incl

Looks like the issue was the `container` in `hash::combine(seed, container)` can't be a ranges::views type. It works when it's a vector or sth. like that. So this patch request can be discarded.

PR #63, opened 2022-02-17
Nakul/production/open shell eval

nakulteke:
All features required for a general order open-shell spin tracing are included here. The anti-symmetrizer from spin-orbital coupled cluster equations is partially expanded, and the corresponding partial antisymmetrizer is used to generate expression compactly.


SeQuant/core/tensor.hpp
```diff
@@ -160,24 +160,33 @@ class Tensor : public Expr, public AbstractTensor {
     return bra_rank();
   }
 
-  /// @Brief Convert Fock and Coulomb integrals to std::wstring formula compatible with MPQC v4
+  /// @Brief Convert Fock and Coulomb integrals to std::wstring formula
```
evaleev, line 163: `@Brief` is not a Doxygen command, `@brief` is

evaleev (approved):

PR #76, opened 2022-05-02
Implements tot evaluation and evaluation order improvements

bimalgaudel:
- Evaluate, particle-symmetrize, and particle-antisymmetrize PNO-like expressions.
- Evaluation order made explicit. Expressions are evaluated from left to right as they appear in `Expr`.
- Bug fixes and refactors.


CMakeLists.txt
```diff
@@ -281,7 +281,7 @@ set_source_files_properties(
         "SEQUANT_GIT_REVISION=\"${SEQUANT_GIT_REVISION}\";SEQUANT_GIT_DESCRIPTION=\"${SEQUANT_GIT_DESCRIPTION}\""
 )
 
-target_link_libraries(SeQuant PUBLIC range-v3::range-v3 Boost::regex Boost::locale Boost::boost SeQuant-bliss Threads::Threads)
```
evaleev: `wstring.hpp` uses Boost.Locale so dependence on it is real ... why remove it?

bimalgaudel: I wonder why the CI runs do not fail at all even when Boost::locale linking is removed.

Anyway, I was trying to compile SeQuant with Boost-1.81(manually installed), Boost::locale was not an importable target. The reason for that could be that I did not pass `--with-locale` flag while compiling Boost.

evaleev: @bimalgaudel SeQuant CI tests never failed (e.g. https://github.com/ValeevGroup/SeQuant/pull/76/commits/a9980513d3b287d53a626e6c13f4e75c3348dded) , the issue is exposed by MPQC CI tests. My guess is SeQuant test coverage is not sufficient (no need to focus on increasing SeQuant coverage necessarily, some testing is only possible at scale within the context of MPQC).

Of course we assume all targets of Boost are available, incl. locale.

PR #95, opened 2023-06-13
[test]: Update Catch to v2.13.9

### LINKED ISSUE #94: Catch2 needs to be updated
SeQuant still uses an older version of Catch2 that is no longer compatible with more recent glibc versions, because they have changed `MINSIGSTKSZ` to no longer be a constant expression in v2.35 (and older Catch2 versions use this variable as a constant expression).

The issue is fixed in Catch v2.13.5

As it stands, this causes `sequant/unit/build` to fail when running the test cases.

Ref.: https://github.com/catchorg/Catch2/issues/2421

Krzmbrzl:
This is the latest release of Catch v2. It contains a fix for MINSIGSTKSZ no longer being a constant expressions.

Fixes #94

### DIFF
--- a/tests/unit/catch.hpp
+++ b/tests/unit/catch.hpp
@@ -1,9 +1,9 @@
 /*
- *  Catch v2.13.4
- *  Generated: 2020-12-29 14:48:00.116107
+ *  Catch v2.13.9
+ *  Generated: 2022-04-12 22:37:23.260201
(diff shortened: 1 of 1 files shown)


evaleev:
@Krzmbrzl thank you for the PR ... apologies for tardy response [vacation]

evaleev (approved):

PR #109, opened 2023-07-05
use multiprecision for `sequant::factorial`

ajay-mk:


SeQuant/core/math.hpp
```diff
@@ -15,30 +16,42 @@ constexpr std::size_t pow2(std::size_t n) {
...
+    return values[n];
+  else {
+    auto fac = [](std::size_t n) {
+      int128_t result = 1;
+      for (auto i = 2; i <= n; ++i) {
+        result *= i;
+      }
+      return result;
+    };
+    return fac(n);
```
Krzmbrzl, line 53: 1. Why define a lamda for calculating the factorial and then immediately calling that instead of simply computing the factorial straight away
2. You don't need to recompute the factorial up to `n==20`:
```suggestion
  int128_t result = values[20];
  for (auto i = 21; i <= n; ++i) {
    result *= i;
  }
```
or even better:
```suggestion
  constexpr auto nPrecomputed = std::size(values);
  int128_t result = values[nPrecomputed - 1];
  for (auto i = nPrecomputed + 1; i <= n; ++i) {
    result *= i;
  }
  return result;
```

examples/stcc/stcc.cpp
```diff
@@ -139,9 +139,15 @@ ExprPtr biorthogonal_transform(
   // Coefficients
   std::vector<double> bt_coeff_vec;
   {
+    auto factorial = [](auto x) {
+      if (x > 20)
+        throw std::runtime_error("std::intmax_t running out of precision");
+      return boost::numeric_cast<std::intmax_t>(sequant::factorial(x));
+    };
```
Krzmbrzl, line 146: Wouldn't it make more sense to use `std::uintmax_t` instead of `std::intmax_t`?

SeQuant/core/math.hpp
evaleev: thanks, addressed, `fac` is gone

examples/stcc/stcc.cpp
evaleev: will address via https://github.com/ValeevGroup/SeQuant/issues/110

PR #117, opened 2023-07-11
gaudel/generalized evaluation

bimalgaudel:
Supports:
- tensor * tensor -> tensor
- tensor * tensor -> scalar


SeQuant/core/eval_expr.hpp
```diff
@@ -1,190 +1,50 @@
...
- */
-class EvalExpr {
- public:
-  using hash_t = size_t;
+enum class EvalOp { Id, Sum, Prod };
+
+enum class ResultType { Tensor, Constant };
 
-  EvalExpr() = delete;
+class EvalExpr {
```
evaleev, line 13: dox

SeQuant/core/eval_expr.hpp
```diff
@@ -1,190 +1,50 @@
...
- *
- * @author Bimal Gaudel
- * @version 03 Dec, 2020
- */
-class EvalExpr {
- public:
-  using hash_t = size_t;
+enum class EvalOp { Id, Sum, Prod };
+
+enum class ResultType { Tensor, Constant };
```
evaleev, line 11: dox

SeQuant/domain/eval/eval_result.hpp
```diff
@@ -0,0 +1,683 @@
...
+}
+
+std::logic_error unimplemented_method(std::string_view msg) noexcept {
+  using namespace std::string_literals;
+  return std::logic_error{"Not implemented in this derived class: "s +
+                          msg.data()};
+}
+
+template <typename T>
+struct Annot {
```
evaleev, line 33: dox

SeQuant/domain/eval/eval_result.hpp
```diff
@@ -0,0 +1,683 @@
...
+
+  symmetrize_backend(rank, groups, call_back);
+
+  TA::DistArray<Args...>::wait_for_lazy_cleanup(result.world());
+
+  return result;
+}
+
+template <typename... Args>
+auto antisymmetrize_ta(
```
evaleev, line 253: dox

SeQuant/domain/eval/eval_result.hpp
```diff
@@ -0,0 +1,683 @@
...
+    // Which is a potentailly narrowing conversion leading to
+    // integral overflow. Hence, the values in the returned
+    // container are mixed negative and positive integers (long type)
+    //
+    return static_cast<long>(sequant::hash::value(idx.label()));
+  });
+}
+
+template <typename... Args>
+auto symmetrize_ta(TA::DistArray<Args...> const& arr,
```
evaleev, line 229: dox

evaleev (changes requested):

SeQuant/domain/eval/eval_result.hpp
```diff
@@ -0,0 +1,684 @@
...
+  template <typename T>
+  [[nodiscard]] T& get_ref() {
+    assert(has_value());
+    return *std::any_cast<T>(&value_);
+  }
+
+  template <typename T,
+            typename = std::enable_if_t<!std::is_convertible_v<T, EvalResult>>>
+  explicit EvalResult(T&& arg) noexcept
+      : value_{std::make_any<T>(std::forward<T>(arg))} {}
```
evaleev, line 441: `T` can be deduced to be a reference type here, need to `make_any<std::decay<T>>(`

SeQuant/core/eval_expr.cpp
```diff
@@ -1,140 +1,217 @@
...
+          remove_item(right, rr);
+          goto next_contract;
+        } else if (lb == rk) {
+          std::swap(lk, rk);
+          remove_item(left, ll);
+          goto next_contract;
+        } else if (lk == rb) {
+          std::swap(lb, rb);
+          remove_item(left, ll);
+          goto next_contract;
```
Krzmbrzl, line 177: To break from a nested loop, you should only jump forward using `goto`, but never backwards: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-goto

SeQuant/core/eval_expr.hpp
```diff
@@ -1,190 +1,50 @@
...
+  size_t id_;
+  ExprPtr expr_;
+  bool tot_;
 
-  /**
-   * Construct EvalExpr that goes into the leaf nodes
-   * of the binary evaluation tree.
-   */
-  explicit EvalExpr(const Tensor&) noexcept;
+  static size_t global_id_;
```
Krzmbrzl, line 22: Typically, class definitions are easier to understand if private members are defined after public ones ('cause most of the time one looks up a class definition, one is interested in how one can interact with the class from the outside.)

SeQuant/core/eval_expr.hpp
```diff
@@ -1,190 +1,50 @@
...
 
- private:
-  using index_container_type = container::svector<Index>;
-  using braket_type = std::pair<index_container_type, index_container_type>;
+  [[nodiscard]] bool tot() const noexcept;
 
-  size_t id_;
+  [[nodiscard]] std::wstring to_latex() const noexcept;
+
+  [[nodiscard]] sequant::Tensor const& as_tensor() const noexcept(false);
```
Krzmbrzl, line 45:
```suggestion
  [[nodiscard]] sequant::Tensor const& as_tensor() const;
```
There is no point in adding `noexcept(false)` to a regular function. The only case where this can make sense is with destructors as these are `noexcept` by default. See also https://stackoverflow.com/q/16244313

SeQuant/core/optimize.hpp
```diff
@@ -100,31 +100,53 @@ struct OptRes {
...
 template <typename I1, typename I2>
-std::pair<container::svector<Index>, container::svector<Index>> common_indices(
-    I1 const& idxs1, I2 const& idxs2) {
+container::svector<Index> common_indices(I1 const& idxs1, I2 const& idxs2) {
   using std::back_inserter;
   using std::begin;
   using std::end;
+  using std::set_intersection;
+
+  auto result = container::svector<Index>{};
```
Krzmbrzl, line 116:
```suggestion
  container::svector<Index> result;
```
More compact (and concise)

SeQuant/core/optimize.hpp
```diff
@@ -100,31 +100,53 @@ struct OptRes {
...
+
+  set_intersection(begin(idxs1), end(idxs1), begin(idxs2), end(idxs2),
+                   back_inserter(result), Index::LabelCompare{});
```
Krzmbrzl, line 119:
```suggestion
  assert(std::is_sorted(begin(idx1), end(idx1), Index::LabelCompare{}));
  assert(std::is_sorted(begin(idx2), end(idx2), Index::LabelCompare{}));
  set_intersection(begin(idxs1), end(idxs1), begin(idxs2), end(idxs2),
                   back_inserter(result), Index::LabelCompare{});
```

SeQuant/core/optimize.hpp
```diff
@@ -100,31 +100,53 @@ struct OptRes {
...
+container::svector<Index> diff_indices(I1 const& idxs1, I2 const& idxs2) {
+  using std::back_inserter;
+  using std::begin;
+  using std::end;
+  using std::set_symmetric_difference;
+
+  auto result = container::svector<Index>{};
+
+  set_symmetric_difference(begin(idxs1), end(idxs1), begin(idxs2), end(idxs2),
+                           back_inserter(result), Index::LabelCompare{});
```
Krzmbrzl, line 138:
```suggestion
  assert(std::is_sorted(begin(idx1), end(idx1), Index::LabelCompare{}));
  assert(std::is_sorted(begin(idx2), end(idx2), Index::LabelCompare{}));
  set_symmetric_difference(begin(idxs1), end(idxs1), begin(idxs2), end(idxs2),
                           back_inserter(result), Index::LabelCompare{});
```

SeQuant/core/optimize.hpp
```diff
@@ -100,31 +100,53 @@ struct OptRes {
...
+/// @note I1 and I2 containers are assumed to be sorted by using
+/// Index::LabelCompare{};
+template <typename I1, typename I2>
+container::svector<Index> diff_indices(I1 const& idxs1, I2 const& idxs2) {
+  using std::back_inserter;
+  using std::begin;
+  using std::end;
+  using std::set_symmetric_difference;
+
+  auto result = container::svector<Index>{};
```
Krzmbrzl, line 135:
```suggestion
  container::svector<Index> result;
```

SeQuant/core/optimize.hpp
```diff
@@ -100,31 +100,53 @@ struct OptRes {
...
+}
+
+///
+/// TODO: Use C++20 <bit> header when possible
+/// \tparam T integral type
+/// \return true if @c x has a single bit on in its bit representation.
+///
+template <typename T>
+bool has_single_bit(T x) noexcept {
+  return x != 0 && (x & (x - 1)) == 0;
```
Krzmbrzl, line 149:
```suggestion
  return std::bitset<sizeof(T)>(x).count() == 1;
```
This is more readable imo (and the compiler should be able to optimize away all overhead this might introduce)

bimalgaudel: There is a function in standard library in C++20. I am trying to emulate that.

https://en.cppreference.com/w/cpp/numeric/has_single_bit

SeQuant/core/eval_expr.cpp
bimalgaudel: This is a tricky situation. Also this function will be generalized for contraction involving tensors with unequal number of bra and ket indices. It can/will be addressed then.

SeQuant/core/optimize.hpp
bimalgaudel: It is a matter of taste. Writing it the way you suggested, the variable name `result` is now drowned in the preceding type specification (imo).

SeQuant/core/optimize.hpp
Krzmbrzl: I know. But writing bit-fiddling down explicitly usually is reasonably unreadable. 
https://quick-bench.com/q/LSJ_Spc0QvfGRECIMV3RWVwalwQ indicates that there are more readable ways to write this down that only intrdoduce a small overhead in performance. Interestingly, C++20's `has_single_bit` doesn't seem to perform that well. Even a simple `x % 2 == 0` performs better :eyes: 
Anyway, I only wanted to point out that there is a small potential for making the code more readable for non-bit-wizards. Feel free to ignore my comment though, if you don't think anything needs to be changed :)

evaleev: using "standard" names is the way to go. Even when the name is eminently confusing (like `has_single_bit`). Unfortunately we are not ready to _require_ C++20 across the stack (SeQuant is a small part of it) ... not yet at least. But it's a good idea to actually use `std::has_single_bit` directly if C++20 is available (`#if __cplusplus >= ...`)

bimalgaudel (commented):
The requests have been addressed.

evaleev (approved):

Issue #97, opened 2023-06-26
Default convention is incomplete (w.r.t. index space types)


Krzmbrzl:
As far as I can tell, the default QCiFS convention doesn't register all possible index space types.

All types are listed as
https://github.com/ValeevGroup/SeQuant/blob/1dd6257ba9e5b2e9aa2b71eb38768f8a8a1699c7/SeQuant/core/space.hpp#L145-L158
whereas the convention (seems to) only defines
https://github.com/ValeevGroup/SeQuant/blob/1dd6257ba9e5b2e9aa2b71eb38768f8a8a1699c7/SeQuant/domain/mbpt/convention.cpp#L75-L92

Ignoring the null/non-null space types, the following spaces are not covered by the default convention:
- frozen_unoccupied
- inactive_occupied
- inactive_unoccupied

Maybe these spaces are not part of the QCiFS paper series and have therefore left out, but from a user's point of view, it would be very beneficial, if the default convention would indeed register all existing index space types instead of only a subset of them.

evaleev:
@Krzmbrzl As part of #112 I documented these choices as well as extended the convention to include active orbitals (https://github.com/ValeevGroup/SeQuant/pull/112/commits/0177f6592ae9f0304d986126c7b8f7aade63c302) ... as is often the case with historical artifacts the QCiFS "convention" (or, rather, what I define as the convention) is not systematic. I would prefer actually to only keep in the convention the symbols that have been established as part of the convention. I have not seen established uses of all "standard" spaces, so I'd actually rather them undefined.

Note that nothing prevents the user from registering additional choices, e.g.
```
IndexSpace::register_instance(L"m'", IndexSpace::frozen_occupied);

// now can use m'
Index m_1(L"m'");
```
(and defining additional IndexSpaces, etc.). If user wants to add something to the convention they need to make a PR that documents the proposed choice.

Krzmbrzl:
> I would prefer actually to only keep in the convention the symbols that have been established as part of the convention. I have not seen established uses of all "standard" spaces, so I'd actually rather them undefined.

I understand.

The issue only comes about when e.g. the wick theorem tries to create an index in the intersection between two pre-defined index spaces but then the program errors because that intersection space has not been given a name.
[...]

Issue #96, opened 2023-06-21
Internal encoding and string type


Krzmbrzl:
From e.g.
https://github.com/ValeevGroup/SeQuant/blob/1dd6257ba9e5b2e9aa2b71eb38768f8a8a1699c7/examples/srcc/srcc.cpp#L97-L98
I tend to imply that SeQuant wants to use UTF-8. However, in the code you chose to use wide strings (`std::wstring`), which would instead indicate that you really want to use UTF-16 or UTF-32.

This opens up two questions for me:
1. What encoding is SeQuant using internally? UTF-8 or UTF-16/UTF-32?
2. If you're indeed using UTF-8: Why are you using wide strings to represent that? Normally, one would use regular strings for UTF-8 (which would also eliminate the need for having to prefix all string literals with `L`)

evaleev:
@Krzmbrzl yes, we use UTF-8 "internally". The decision to use `std::wstring` was indeed made a long ago when I did not necessarily understand subtleties of unicode processing (we've been using `std::wstring` in MPQC since early 2010s). Our use of strings in SeQuant is primarily for _character_-centric encoding of data structures; viewing the string as a sequence of characters should be as easy/familiar to non-specialists as possible; wide characters make it natural. With `char`-based encoding asking questions like "does this `Index` contain single label or subscripted with a numerical index" becomes far more involved.

Clearly you are concerned about `std::wstring` leaking out the API ... how much of a deal-breaker is this?

Krzmbrzl:
> Clearly you are concerned about std::wstring leaking out the API ... how much of a deal-breaker is this?

I wouldn't exactly call it a deal-breaker. From my experience, it's just a little unusual to be required to use wide strings and wide characters for everything. Especially since in SeQuant most strings are tensor and index names, which generally tend to be rather short. Thus, the required additional `L` prefix in front of every literal has a relatively high overhead in terms of characters that need to be typed in order to get things done.
But this is really just a bit of an inconvenience and not a deal-breaker.

> Our use of strings in SeQuant is primarily for character-centric encoding of data structures; viewing the string as a sequence of characters should be as easy/familiar to non-specialists as possible; wide characters make it natural. 

Yes and no. Based on my research on wide character / wide string support in C++, it seems like the size of a wide character is not the same across different platforms. E.g. on Windows it seems to be only 2 bytes wide, which means that BMP characters can be represented in one code point (e.g. one wide character), but any non-BMP character will still use multiple code points per character.

[...]

PR #124, opened 2023-07-27
Improve missing space instance error msg

Krzmbrzl:
Instead of only throwing a bad_attr exception without any (useful) error message, we now construct a string representation of the requested space type and quantum numbers.
This should make debugging missing-instance errors more easily.

Step towards #97

### DIFF
--- a/SeQuant/core/space.hpp
+++ b/SeQuant/core/space.hpp
@@ -7,6 +7,8 @@
 
 #include <bitset>
 #include <cassert>
+#include <codecvt>
+#include <locale>
 
 #include "attr.hpp"
 #include "container.hpp"
@@ -107,6 +109,9 @@ struct QuantumNumbersAttr {
     return QuantumNumbersAttr(-0);
   }
 };
+
+std::wstring to_wstring(TypeAttr type);
+std::wstring to_wstring(QuantumNumbersAttr qns);
 
 /// @brief space of Index objects
 ///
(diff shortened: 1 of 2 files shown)


SeQuant/core/space.hpp
```diff
@@ -340,7 +346,12 @@ class IndexSpace {
     const auto attr = Attr(type, qns);
     assert(attr.is_valid());
     if (attr == Attr::null()) return null_instance();
-    if (!instance_exists(attr)) throw bad_attr();
+    if (!instance_exists(attr)) {
+      std::wstring_convert<std::codecvt_utf8_utf16<wchar_t>> converter;
+      throw bad_attr(converter.to_bytes(L"Request to non-existing space: " +
```
Krzmbrzl, line 351: Maybe
```suggestion
      throw bad_attr(converter.to_bytes(L"Request to non-registered space: " +
```
is preferable?

Krzmbrzl:
@evaleev is there anything I still need to do for this one to be merged?

SeQuant/core/space.hpp
Krzmbrzl: To my knowledge they have first deprecated and now apparently removed the `codecvt` header in the C++ standard but without offering any alternative. That means one would have to write custom platform-depend code to perform these conversions - see also https://stackoverflow.com/a/42946556

Maybe another reason to reconsider #96? :D

PR #129, opened 2023-08-02
Unicode characters (non-space) can appear as tensor label while parsing string into sequant expression.

bimalgaudel:


evaleev:
I'm not sure if this is a good idea after all ... are we supposed to be doing the reverse of utf_to_latex here? Otherwise latex -> SQ -> latex will not roundtrip ...

bimalgaudel:
@evaleev parse_expr function does not take LaTeX. It takes a string that has some pattern, a tiny DSL and spits out sequant expression if the string is valid. There is a deparse_expr function that does the reverse: takes a sequant expression and produces a string that can be parsed into the same expression using parse_expr.

The tensor labels are not supposed to be in LaTeX form, rather they should be what the Tensor.label() method should return. So it makes sense to support unicode labels since we are already using them: t¹, λ etc.

evaleev:
Ok thanks for the explainer ... I thought the dsl was latex.

PR #134, opened 2023-08-12
Compute minimum storage required for the most memory intensive of evaluation.

bimalgaudel:


SeQuant/core/asy_cost.cpp
```diff
@@ -202,6 +215,14 @@ bool operator>(AsyCost const &lhs, AsyCost const &rhs) {
   return !(lhs < rhs || lhs == rhs);
 }
 
+bool operator<=(AsyCost const& lhs, AsyCost const& rhs) {
+  return lhs < rhs || lhs == rhs;
```
Krzmbrzl, line 219:
```suggestion
  return !(lhs > rhs);
```
(tiny bit more efficient)

SeQuant/core/asy_cost.cpp
```diff
@@ -202,6 +215,14 @@ bool operator>(AsyCost const &lhs, AsyCost const &rhs) {
...
+}
+
+bool operator>=(AsyCost const& lhs, AsyCost const& rhs) {
+  return lhs > rhs || lhs == rhs;
```
Krzmbrzl, line 223:
```suggestion
  return !(lhs < rhs);
```

SeQuant/core/binary_node.hpp
```diff
@@ -9,6 +10,77 @@
...
+/// Visit children nodes followed by parent node.
+/// Left will be visited before right.
+struct PreOrder {};
+
+/// Visit parent node followed by children nodes.
+/// Left will be visited before right.
+struct PostOrder {};
+
+/// Visit left node then parent node followed by right node.
+struct InOrder {};
```
Krzmbrzl, line 27: Might it make sense to group these into an enum?
```suggestion
enum class IterationOrder {
  Pre,
  Post,
  In,
};
```

bimalgaudel: Then we cannot use tag dispatch. Which I intend to use.

Krzmbrzl: Ah, I see.

One could `if constexpr` on the enum value and call the desired overloads, but I guess that could get clumsy :eyes:

PR #135, opened 2023-08-13
improved spin-free closed-shell biorthogonalization

evaleev:
Current biorothogonalization using the pseudoinverse of the excitation manifold metric does not reduce the number of terms for CCSDT and higher. The new approach using the metric QR should be an improvement,


Krzmbrzl:
> The new approach using the metric QR should be an improvement

Is this effectively the approach described in [this work](https://doi.org/10.48550/arXiv.1805.00565) or is this a different approach?

The method described in that paper creates a "semi-biorthogonal" basis (in their approach obtained by a least-squares minimization) and then removes a lot of terms by projecting out all residual tensor components within the null-space of the overlap matrix of the configurations generated by the $\hat{E}$ operators.

evaleev:
@Krzmbrzl yes along the lines of Gerald's work, but improved ... the equations published on their group's site are not as compact as they can be

evaleev:
@Krzmbrzl if you have better approaches in my mind I'm all ears ... spin-adaptation of CCSDT is not described well in the literature and it seems to generally involve relatively sophisticated representation theory ...

Krzmbrzl:
Nah, I don't have better ideas. I was simply looking into the subject and was wondering whether you are implementing an approach that I have not yet heard of.
Maybe we will eventually look into this subject again for our icMRCC equations, but for now I have to get them generated in the first place :D

PR #136, opened 2023-08-23
Support energy expressions to be spintraced

Krzmbrzl:
The closed_shell_CC_spintrace function would error if it was fed the CC energy equations, because it unconditionally performs a biorthogonal transform, which assumes the presence of external indices (aka: it assumes that we are dealing with a CC residual equation).

This commit changes the code so that we only perform the transformation for residuals, but skip it for energy expressions (fully contracted expressions).

### DIFF
--- a/tests/unit/test_spin.cpp
+++ b/tests/unit/test_spin.cpp
@@ -663,20 +663,14 @@ SECTION("Closed-shell spintrace CCD") {
     const auto input = ex<Sum>(ExprPtrList{parse_expr(
         L"1/4 g{i_1,i_2;a_1,a_2} t{a_1,a_2;i_1,i_2}", Symmetry::antisymm)});
     auto result = closed_shell_CC_spintrace(input);
-
-	// Canonicalize the order of the addends in the result
-	ExprPtr firstAddend;
-	ExprPtr secondAddend;
-	if (result.as<Sum>().summand(0).as<Product>().scalar() == -1) {
-		firstAddend = result.as<Sum>().summand(1);
-		secondAddend = result.as<Sum>().summand(0);
-	} else {
-		firstAddend = result.as<Sum>().summand(0);
-		secondAddend = result.as<Sum>().summand(1);
-	}
-
-	REQUIRE(*firstAddend == *parse_expr(L"2 g{i_1,i_2;a_1,a_2} t{a_1,a_2;i_1,i_2}", Symmetry::nonsymm));
-	REQUIRE(*secondAddend == *parse_expr(L"-g{i_1,i_2;a_1,a_2} t{a_1,a_2;i_2,i_1}", Symmetry::nonsymm));
+    if constexpr (hash_version() == hash::Impl::BoostPre181)
+      REQUIRE(result == parse_expr(L"- g{i_1,i_2;a_1,a_2} t{a_1,a_2;i_2,i_1} + "
+                                   L"2 g{i_1,i_2;a_1,a_2} t{a_1,a_2;i_1,i_2}",
+                                   Symmetry::nonsymm));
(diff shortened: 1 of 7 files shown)


bimalgaudel (approved):

tests/unit/test_spin.cpp
```diff
@@ -655,6 +657,18 @@ SECTION("Swap bra kets") {
   }
 }
 
+SECTION("Closed-shell spintrace CCD") {
+  // Energy expression
+  {
+    const auto input = ex<Sum>(ExprPtrList{parse_expr(
+        L"1/4 g{i_1,i_2;a_1,a_2} t{a_1,a_2;i_1,i_2}", Symmetry::antisymm)});
+    auto result = closed_shell_CC_spintrace(input);
+    REQUIRE(*result == *parse_expr(L"-g{i_1,i_2;a_1,a_2} t{a_1,a_2;i_2,i_1} + "
```
evaleev, line 666: this check seems to be non-portable, e.g. see https://github.com/ValeevGroup/SeQuant/actions/runs/6811607068/job/18522290446?pr=136#step:10:291

evaleev (changes requested):

tests/unit/test_spin.cpp
Krzmbrzl: Yeah the ordering of the addends in the result seems to be in reverse order on the CI compared to my local setup. No idea why that would happen, but I have added an explicit ordering step to the test case and now it succeeds.

evaleev (approved):

tests/unit/test_spin.cpp
evaleev: Canonicalizer output unfortunately depends on the Boost.Hash version and they changed how hashing works in 1.81 ... I don't want to reimplement hashing (and Boost.Hash > Std.Hash) so for some unit tests we have to provide both variants

Krzmbrzl:
@evaleev what's the reasoning behind [7d2b9e8](https://github.com/ValeevGroup/SeQuant/pull/136/commits/7d2b9e8f07fbac9b15435c325a53813a7f1e9306)? Normally, without dereferencing the comparison would be on pointer-equality, but due to
https://github.com/ValeevGroup/SeQuant/blob/9c7a2225fb1632b72caea7d25000dbef7b1078b4/SeQuant/core/expr_operator.hpp#L10-L12
this turns out to be a comparison on the dereferenced values anyway. So functionally the old code and the new one do the same thing, don't they? Does that mean that your changes were merely a cosmetic change? Or is there some functionality/intricacy that I have missed? :thinking:

evaleev:
> this turns out to be a comparison on the dereferenced values anyway. So functionally the old code and the new one do the same thing, don't they? Does that mean that your changes were merely a cosmetic change? Or is there some functionality/intricacy that I have missed? 🤔

Purely cosmetic, saving 2 characters. I saw that you followed the compare-after-deref and wanted to fix the original examples. Since now (as of ~6 month ago) `ExprPtr` is a proper class (rather than alias to `shared_ptr<Expr>`) derefing `ExprPtr` is not needed.

bimalgaudel:
Also, now that `ExprPtr` is a 'proper' [class](https://github.com/ValeevGroup/SeQuant/blob/9c7a2225fb1632b72caea7d25000dbef7b1078b4/SeQuant/core/expr.hpp#L37) rather than an alias to `std::shared_ptr<..>`, `ExprPtr` should be treated as objects rather than pointers when possible, I think. :)

Krzmbrzl:
Ah, I always thought they were merely an alias for a `shared_ptr`, but them being their own class names a couples of things more clear ^^

PR #138, opened 2023-08-23
Revamp and fix enable-ifs in expr.cpp

Krzmbrzl:
The used enable-ifs looked a bit scary and have therefore been
refactored to use intermediates with (hopefully) clear names.

Furthermore, a couple of tests in those enable-ifs have been performed
with std::is_same, but only removing const from the compared type and
thus these checks would not work properly as soon as a reference type
was passed.
This could be observed while constructing a Product from a non-const
other product. Thus, the type of the to-be-copied object was just
Product & and not const Product &, which is what the copy-ctor
(correctly) is defined for. However, there is the templated
Product(Range &&) ctor, that thanks to the universal reference is the
more specialized candidate in this case (it takes the type as-is without
requiring a const to be slapped on). This would lead to the product
being copied, but also flattened (as this is the default behavior of the
range-ctor).

### DIFF
--- a/SeQuant/core/expr.hpp
+++ b/SeQuant/core/expr.hpp
@@ -14,6 +14,7 @@
 #include "SeQuant/core/logger.hpp"
 #include "SeQuant/core/meta.hpp"
 #include "SeQuant/core/rational.hpp"
+#include "SeQuant/core/meta.hpp"
 #include "SeQuant/core/wolfram.hpp"
 
 #include <range/v3/all.hpp>
@@ -26,9 +27,40 @@
 #include <iostream>
 #include <memory>
 #include <optional>
+#include <type_traits>
 #include <vector>
 
 namespace sequant {
+
+namespace {
+
+
+template <typename T>
+constexpr bool is_an_expr_v = meta::is_base_of_v<Expr, T>;
+template <typename T>
+constexpr bool is_expr_v = meta::is_same_v<Expr, T>;
+
+template <typename T>
+constexpr bool is_a_constant_v = meta::is_base_of_v<Constant, T>;
+template <typename T>
+constexpr bool is_constant_v = meta::is_same_v<Constant, T>;
+
+template <typename T>
+constexpr bool is_a_sum_v = meta::is_base_of_v<Sum, T>;
+template <typename T>
+constexpr bool is_sum_v = meta::is_same_v<Sum, T>;
+
+template <typename T>
+constexpr bool is_a_product_v = meta::is_base_of_v<Product, T>;
+template <typename T>
+constexpr bool is_product_v = meta::is_same_v<Product, T>;
+
+template <typename T>
+constexpr bool is_a_variable_v = meta::is_base_of_v<Variable, T>;
+template <typename T>
+constexpr bool is_variable_v = meta::is_same_v<Variable, T>;
+
+}  // namespace
 
 /// @brief ExprPtr is a multiple-owner smart pointer to Expr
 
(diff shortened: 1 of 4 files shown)


SeQuant/core/expr.hpp
```diff
@@ -369,25 +392,25 @@ class Expr : public std::enable_shared_from_this<Expr>,
   /// @return true if this object is of type @c T
   template <typename T>
   bool is() const {
-    if constexpr (std::is_same_v<std::decay_t<T>, Expr>)
+    if constexpr (is_expr_v<T>)
       return true;
     else
-      return this->type_id() == get_type_id<std::decay_t<T>>();
+      return this->type_id() == get_type_id<remove_cvref_t<T>>();
```
Krzmbrzl, line 398: There is a subtle difference between `remove_cvref ` and `decay` - see https://stackoverflow.com/a/74257596/3907364
But I think in this particular case, it doesn't matter (and since decay is not needed, I replaced it with the "less specialized" remove_cvref)

SeQuant/core/expr.hpp
```diff
@@ -369,25 +392,25 @@ class Expr : public std::enable_shared_from_this<Expr>,
...
   }
 
   /// @tparam T an Expr type
   /// @return this object cast to type @c T
   template <typename T>
   const T &as() const {
-    assert(this->is<std::decay_t<T>>());  // so that as<const T>() works fine
+    assert(this->is<T>());
```
Krzmbrzl, line 405: Note that the `is<T>()` will perform the cv_ref removal, so it is not needed here

SeQuant/core/type_traits.hpp
Krzmbrzl, line 1: Maybe these should instead go into `meta.hpp`? I'm not sure if type-traits have to fulfill some property in order to go there...

Krzmbrzl:
Rebased against current master

SeQuant/core/type_traits.hpp
```diff
@@ -0,0 +1,47 @@
+//
```
evaleev, line 1: meta programming (including type traits) are already present in `meta.hpp`, should go there, unless there is a good reason. not to. Also, to be "safe" meta programming lives in `sequent::meta` namespace since most of the time users do not need to bother

SeQuant/core/type_traits.hpp
```diff
@@ -0,0 +1,47 @@
...
+/// ignoring CV and reference qualifiers
+template <typename Base, typename T>
+using is_a = std::is_base_of<remove_cvref_t<Base>, remove_cvref_t<T>>;
+template <typename Base, typename T>
+constexpr bool is_a_v = is_a<Base, T>::value;
+
+/// Checks whether \c T and \c U are the same type, ignoring any CV and
+/// reference qualifiers
+template <typename T, typename U>
+struct is
```
evaleev, line 38: ultra-short names scare me ... how about `sequent::meta_is_same{,_v}` for this? This is basically a slightly improved version of `std::is_same{,_v}`, so makes sense to keep the name.

in same vein: `is_base_of{,_v}` preferred to `is_a{,_v}`

evaleev (changes requested):

SeQuant/core/type_traits.hpp
Krzmbrzl: I shied away from using longer names as they are essentially exclusively used in nested template specifiers and these get immensely long already...
But if you prefer, I can definitely adapt the naming and try my best at formatting the resulting template code.

Krzmbrzl:
Addressed your review comments and rebased against current `master`

PR #140, opened 2023-08-24
add complete_inactive_unoccupied space.

cmasteran1:
add complete_inactive_unoccupied space as a union of other_unoccupied and inactive_unoccupied for CT-F12 methods in MPQC

### DIFF
--- a/SeQuant/core/space.hpp
+++ b/SeQuant/core/space.hpp
@@ -242,7 +242,7 @@ class IndexSpace {
   static constexpr Type other_unoccupied{0b1000000};
   /// a union of the other_unoccupied space and inactive_unoccupied space.
   /// @note useful when considering active unoccupied orbitals separately from frozen orbitals in F12 theory
-  static constexpr Type complete_inactive_unoccupied{0b1100000};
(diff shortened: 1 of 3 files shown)


SeQuant/core/space.hpp
```diff
@@ -240,6 +240,9 @@ class IndexSpace {
   /// space of sp states that are not used to define the reference (vacuum) state (i.e., they are unoccupied) and not supported
   /// by a supported by a finite computational basis; i.e., these states are the rest of the sp Hilbert space
   static constexpr Type other_unoccupied{0b1000000};
+  /// a union of the other_unoccupied space and inactive_unoccupied space.
+  /// @note useful when considering active unoccupied orbitals separately from frozen orbitals in F12 theory
+  static constexpr Type complete_inactive_unoccupied{0b1100000};
```
Krzmbrzl, line 245:
```suggestion
  static constexpr Type complete_inactive_unoccupied = IndexSpace::other_unoccupied.unIon(IndexSpace::inactive_unoccupied);
```
(that way it is less likely to break should the values of the underlying space types ever be changed)

PR #144, opened 2023-09-13
Conner/feature/fast e3 decomp

cmasteran1:
E3 operators computed via direct substitution instead of applying antisymmetrization procedure.


SeQuant/domain/mbpt/convention.cpp
```diff
@@ -105,6 +107,9 @@ void register_standard_instances() {
     // introduced in MPQC for GF, CT-F12, and other ad hoc uses
     IndexSpace::register_instance(declab(L"x"), IndexSpace::all_active, qnattr,
                                   do_not_throw);
+    // used in virtual projection CT-F12 theory methods in MPQC.
+    IndexSpace::register_instance(declab(L"c"), IndexSpace::complete_inactive_unoccupied, qnattr,
```
evaleev, line 111: too unsystematic ... should use greek letter at a minimum ... how about \gamma (related to g)

evaleev (approved):

evaleev:
- can `antisymmetrizer.hpp:num_closed_loops()` be replaced by `spin.hpp:count_cycles()`
- can `antisymmetrizer.hpp:spin_sum()` be replaced by `spin.hpp:closed_shell_spintrace()`

PR #157, opened 2023-11-15
Feature: Distinguish adjoints with labels

ajay-mk:
This PR introduces the ability to distinguish objects (Tensor and mbpt::Operator) and their adjoints with labels. Unicode superscript character `⁺` (`U+207A`) is added as a suffix to the base label.

### DIFF
--- a/SeQuant/core/tensor.cpp
+++ b/SeQuant/core/tensor.cpp
@@ -12,7 +12,19 @@ void Tensor::assert_nonreserved_label(std::wstring_view label) const {
   // assert(label != overlap_label());
 }
 
+void Tensor::toggle_adjoint_label() {
+  if (is_adjoint_) {
+    if (label_.back() != sequant::adjoint_label)
+      label_.push_back(sequant::adjoint_label);
+  } else {
+    if (label_.back() == sequant::adjoint_label) label_.pop_back();
+  }
+}
+
 void Tensor::adjoint() {
+  // toggle adjoint flag and update label
+  is_adjoint_ = !is_adjoint_;
+  toggle_adjoint_label();
   std::swap(bra_, ket_);
   reset_hash_value();
 }
--- a/SeQuant/core/tensor.cpp
+++ b/SeQuant/core/tensor.cpp
@@ -14,13 +14,19 @@ void Tensor::assert_nonreserved_label(std::wstring_view label) const {
 
 void Tensor::adjoint() {
   std::swap(bra_, ket_);
-  // update label
-  if (!is_adjoint_ && label_.back() != sequant::adjoint_label)
-    label_.push_back(sequant::adjoint_label);
-  else if (is_adjoint_ && label_.back() == sequant::adjoint_label)
-    label_.pop_back();
-  // toggle adjoint flag
-  is_adjoint_ = !is_adjoint_;
+
+  // only track adjointness for BraKetSymmetry::nonsymm cases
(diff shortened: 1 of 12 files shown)


Krzmbrzl:
I would instead like to propose an explicit boolean flag (property) for tensor objects. That's much easier to work with in-code. And for string representation you can still format it however you please.

evaleev:
@Krzmbrzl adjointness is a sticky tag and is part of the state ... I'm still thinking through how this is supposed to interact with the BraKetSymmetry

Krzmbrzl:
For tensors I don't actually see the need for explicit labelling of an adjoint as it should simply be the same tensor but with bra and ket switched, no? And that we can represent immediately: Just swap bra and ket. If the tensor has bra-ket symmetry then canonicalization should swap that back again anyway (or in fact applying the adjoint operation to the tensor could be a no-op).

But even if I am missing something and we really have to/want to explicitly track whether a tensor is an adjoint or not, I think making this property a boolean flag of the tensor object would make working with this property a lot easier and also more performant as it wouldn't require string parsing/manipulation :shrug:

evaleev:
> For tensors I don't actually see the need for explicit labelling of an adjoint as it should simply be the same tensor but with bra and ket switched, no? And that we can represent immediately: Just swap bra and ket. If the tensor has bra-ket symmetry then canonicalization should swap that back again anyway (or in fact applying the adjoint operation to the tensor could be a no-op).

Unfortunately for tensors that are not symmetric/"hermitian" this is necessary. E.g. matrix elements of nonhermitian operators (Hbar, or some F12 formalisms). In practice almost any tensor network is nonhermitian, so representing it by a tensor will necessitate the need to keep track of adjointness, e.g. the V intermediate from F12 theory:

V^{p1p2}_{q1q2} = g_{q1q2}^{a1a2} r_{a1a2}^{p1p2}.

@cmasteran1 learned the importance of distinguishing V and its adjoint the hard way :)

> But even if I am missing something and we really have to/want to explicitly track whether a tensor is an adjoint or not, I think making this property a boolean flag of the tensor object would make working with this property a lot easier and also more performant as it wouldn't require string parsing/manipulation 🤷

Yes, as I mentioned, this is a "sticky" flag, part of `Tensor`'s state.

@ajay-mk I think for a `Tensor` whose `braket_symmetry()` is `symm` or `conjugate` there is no need to update `adjoint_` flag; only for `nonsymm` case `adjoint_` needs to be tracked.

SeQuant/core/tensor.cpp
bimalgaudel, line 1:
```
void Tensor::adjoint() {
		
  std::swap(bra_, ket_);
  if (label_.back() == sequant::adjoint_label) {
    assert(is_adjoint_);
    label_.pop_back();
  } else {
    assert(!is_adjoint_);
    label_.push_back(sequant::adjoint_label);
  }
		
  is_adjoint_ = !is_adjoint_;
		
  reset_hash_value();
		
}
```

SeQuant/domain/mbpt/op.cpp
bimalgaudel, line 1: `to_latex` function need to be this complex?

evaleev: what do you suggest?

ajay-mk:
@evaleev I had to adjust the canonicalize function in TensorNetwork. This should be ready to go now.

evaleev (approved):

PR #160, opened 2023-11-23
Implement ITF translator

Krzmbrzl:
This commit implements functionality that allows to translate a SeQuant expression tree to executable ITF code. The assumption being that said expression tree has been factorized into a binary tree first.

The Integrated Tensor Framework (ITF) is a domain-specific language (DSL) for writing down tensor contractions. The code is interpreted by the Molpro quantum chemistry program package.
For more information see

H.-J. Werner et al., WIREs Comput. Mol. Sci., Vol. 2, No. 2 p. 242-253 (https://doi.org/10.1002/wcms.82)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -190,6 +190,8 @@ set(SeQuant_src
         SeQuant/core/eval_expr.cpp
         SeQuant/core/eval_expr.hpp
         SeQuant/core/eval_node.hpp
+        SeQuant/core/export/itf.cpp
+        SeQuant/core/export/itf.hpp
         SeQuant/core/expr.cpp
         SeQuant/core/expr.hpp
         SeQuant/core/expr_algorithm.hpp
@@ -227,6 +229,7 @@ set(SeQuant_src
         SeQuant/core/tensor_network.hpp
         SeQuant/core/timer.hpp
         SeQuant/core/utility/context.hpp
+        SeQuant/core/utility/indices.hpp
         SeQuant/core/utility/macros.hpp
         SeQuant/core/utility/nodiscard.hpp
         SeQuant/core/utility/singleton.hpp
@@ -422,6 +425,8 @@ if (BUILD_TESTING)
             tests/unit/test_math.cpp
             tests/unit/test_string.cpp
             tests/unit/test_latex.cpp
+            tests/unit/test_utilities.cpp
+            tests/unit/test_export.cpp
             )
 
     if (TARGET tiledarray)
--- a/SeQuant/core/export/itf.cpp
+++ b/SeQuant/core/export/itf.cpp
@@ -1,0 +1,620 @@
+//
+// Created by Robert Adam on 2023-09-04
+//
+
+#include "itf.hpp"
+
+#include <SeQuant/core/utility/indices.hpp>
+
+#include <algorithm>
+#include <array>
+#include <cassert>
+#include <cwchar>
+#include <functional>
+#include <map>
+#include <unordered_map>
+
+namespace sequant {
+
+std::wstring to_itf(const itf::CodeBlock &block) {
+  itf::detail::ITFGenerator generator;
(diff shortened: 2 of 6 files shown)


evaleev:
@Krzmbrzl what you probably want in addition to what you added here is ITF-based evaluation backend. SeQuant's "factorization" capability does not produce sequences of SeQuant expressions since the latter still would need additional work to understand lifetime of intermediates etc. Instead, there is an algorithm that can be customized that will not only optimize/factorize expressions but also "automatically" manage intermediate lifetime. To dump out factorized expressions we need a "dummy" backend that generates SeQuant expressions instead of invoking tensor engine. Currently there is no such backend, but presumably this is something that needs to be added for the ITF converter to be useful ... @bimalgaudel , thoughts?

Krzmbrzl:
Thanks for sharing your thoughts on this.

> SeQuant's "factorization" capability does not produce sequences of SeQuant expressions since the latter still would need additional work to understand lifetime of intermediates etc. Instead, there is an algorithm that can be customized that will not only optimize/factorize expressions but also "automatically" manage intermediate lifetime.

Ah that indeed sounds very useful. Why isn't the optimization & factorization not part of the regular `optimize` function though? Aka: why the need for an evaluation backend in order to perform such optimizations?

Or maybe phrased differently: would it perhaps make sense to make `optimize` return a binary tree (currently called an evaluation tree) immediately?

> it may not be possible to integrate SeQuant into Molpro/ITF or vice versa

Indeed, I think this would be quite complicated to achieve. But we really don't need this at all. We only have to generate ITF code and this can then be interpreted by Molpro as-is. Besides, I expect that as soon as we start implementing icMRCC methods, running SeQuant might take long enough that we wouldn't always want to run SeQuant before being able to perform a computation.

bimalgaudel:
@Krzmbrzl The `optimize` function determines the optimal sequence of evaluation of a term, then reorders the input sequence and puts parentheses. For evaluation the parenthesized `Expr` is converted to a binary evaluation tree using [this function](https://github.com/ValeevGroup/SeQuant/blob/7f4b568f601034cc4c615965f80ff68800164e59/SeQuant/core/eval_node.hpp#L87).

Krzmbrzl:
> For evaluation the parenthesized Expr is converted to a binary evaluation tree using [this function](https://github.com/ValeevGroup/SeQuant/blob/7f4b568f601034cc4c615965f80ff68800164e59/SeQuant/core/eval_node.hpp#L87).

Yeah, but why does this step involve extra optimization in terms of factorization that is not done in `optimize`. It seems to me that reusing intermediates could equally well be formulated by factoring those out in the equations and thus it should be possible to represent expressions that have been factorized this way as an `Expr` object.
So what I am asking is: was this a deliberate choice due to some restrictions that I am currently now aware of or is this something that has historically grown to be done that way? And in the latter case, is there anything speaking against making the extra factorization part of what `optimize` does?

bimalgaudel:
> Yeah, but why does this step involve extra optimization in terms of factorization that is not done in `optimize`.

This step does not involve factorization.

evaleev:
> Yeah, but why does this step involve extra optimization in terms of factorization that is not done in `optimize`. It seems to me that reusing intermediates could equally well be formulated by factoring those out in the equations and thus it should be possible to represent expressions that have been factorized this way as an `Expr` object. So what I am asking is: was this a deliberate choice due to some restrictions that I am currently now aware of or is this something that has historically grown to be done that way? And in the latter case, is there anything speaking against making the extra factorization part of what `optimize` does?

doing common subexpression elimination or other types of optimization will convert a tree to a graph (e.g. a given subexpressions could have multiple consumers/"parents"). Technically `Expr` can represent a graph (since most of the time it uses managed pointers to subexpressions) but this is not recommended. Besides, representing a graph and specifying evaluation sequence as well as managing lifetime of intermediates are different things.

PR #163, opened 2023-12-01
resolve compilation issues in ast.hpp due to cyclic dep

evaleev:


Krzmbrzl:
Where was the cyclic dependence and how does this solve it? Tome the diff looks like your commit mainly makes the constructors explicit :thinking:

evaleev:
> Where was the cyclic dependence and how does this solve it? Tome the diff looks like your commit mainly makes the constructors explicit 🤔

Definition `NullaryValue` was required to instantiate the Product ctor but `NullaryValue` was defined in terms of `Sum` so its definition was needed. Of course `Sum` also requires `Product`

evaleev:
not sure why C++20 triggered the complaint. Indeed might be a compiler issue actually

PR #165, opened 2023-12-06
Conner/feature/exclusionary or spaces

cmasteran1:
User can access excluded spaces between two spaces. Spaces which are not part of the overlap but part of one or the other. At the Attr level this is as simple as applying the xor bitwise operation.


evaleev:
proper term is https://en.wikipedia.org/wiki/Exclusive_or

evaleev:
superseded by https://github.com/ValeevGroup/SeQuant/pull/187

PR #167, opened 2023-12-08
Conner/rework/context dependent spaces

### LINKED ISSUE #123: Switch to run-time index space types
Right now, the different index space types are hard-coded in the source code and those definitions seem to try to cover all (reasonable) use-cases. However, what it all boils down to (it seems) is that the program is able to determine whether or not two given index spaces overlap and if they do, what the overlapping part is.
This is currently done quite conveniently by using bit-masks, but I don't think that it is necessary to define these bitmasks at compile-time.

Instead, I propose letting users define their own index spaces at runtime, without having to use the pre-defined ones which probably contain many spaces that the user isn't actually interested in, but they still have to wrap their head around them to figure out what they are for and whether or not they might be used internally.

If users could simply define their own index spaces and then be sure that the program never uses any other spaces, this would make adoption of SeQuant easier, simply because users can define the environment in the way that they are used to work with and be done with it.

And instead of defining certain checks based on a space's pre-defined type, one could instead work with properties that could be stored as member variables for the individual spaces (or stored somewhere when registering a given space).
For instance, to query whether or not a given index space is occupied in the chosen reference, one could add 
```cpp
enum class Occupation {
	Doubly,
	Never,
	Partially,
};
```
and add a `get_reference_occupation` function to the `Space` class.

This would also alleviate the need to pre-define index spaces for different contexts all at once as this seems to make things a bit more complicated than having dedicated index spaces per application.

cmasteran1:
This change is to help distinguish multi-reference from single reference IndexSpaces. Current master does not produce logical results for some binary set operations because it needs to accommodate partially occupied active space in all contexts. This solution is to make the multi-reference representation orthogonal to the single-reference representation. Crucially this does not add any additional space labels to the standard register since the register is crowded as is. This change is an attempt to resolve #123.


evaleev:
superseded by #187

PR #177, opened 2024-02-05
Implement support for general (auxiliary) tensor indices

### LINKED ISSUE #107: General tensor indices
`Tensor` objects (or more generally: anything fulfilling the `AbstractTensor` concept) are expected to divide their indices into bra and ket indices. There seems to be a connection (at least notation-wise) that bra indices correspond to creator indices and ket indices correspond to annihilator indices.

However, there are situations in which one would want to introduce a tensor object that has indices that are not connected to either a creator or annihilator operator (and would thus neither really fit into the bra or the ket group). Examples of where such tensors may enter are tensor decompositions like e.g. density fitting (resolution of the identity):
$g^{pq}_{rs} \approx B^p_r(Q) B^q_s(Q) - B^p_s(Q) B^q_r(Q)$
where `B` are the DF tensors and `Q` is an index in the auxiliary basis that is used for the DF approximation. This kind of decomposition is only performed on the actual (numeric) tensor object and therefore never appears in the second-quantized formalism and thus is not associated with a creator or annihilator.

While to my understanding the assignment of indices in tensor objects is only for notational purposes (when e.g. exporting to LaTeX), it'd be nice to conceptually keep the creator/annihilator association of tensor indices.

Therefore, I am proposing to extend the tensor concept to support general indices (neither creator nor annihilator).

### LINKED ISSUE #161: Simplify can end up renaming external indices
The following code snippet 
```cpp
auto input = parse_expr(L"-1 S{a_1,a_2;i_1,i_2}:N * f{i_3;i_2}:N * t{a_1,a_2;i_1,i_3}:N");
std::wcout << deparse_expr(input) << "\n";
auto output = simplify(input);

std::wcout << deparse_expr(input) << "\n" << deparse_expr(output) << "\n";
```
produces
```
-1 S{a_1,a_2;i_1,i_2}:N * f{i_3;i_2}:N * t{a_1,a_2;i_1,i_3}:N
-1 S{a_1,a_2;i_1,i_3}:N * f{i_2;i_3}:N * t{a_1,a_2;i_1,i_2}:N
-1 S{a_1,a_2;i_1,i_3}:N * f{i_2;i_3}:N * t{a_1,a_2;i_1,i_2}:N
```
Note that simplification has renamed the external index `i_2` to `i_3` and now uses `i_2` as a contraction index.

Also, I am not sure if calling `simplify` on an expression is supposed to modify the handed-in expression in-place instead of only returning a simplified expression.

Krzmbrzl:
This PR
- adds a new index group (`auxiliary`) to the tensor interface. These indices are (for now) assumed to never carry spin (they are ignored for e.g. spintracing) and they are assumed to be fully non-symmetric.
- Refactor the `TensorNetwork` implementation to make the implementation (hopefully) a bit more readable and also to fix some issues in the old implementation (under certain circumstances the resulting output wasn't canonical). Furthermore, the new implementation ensures that canonicalization is a single-pass step. No more need to sandwich a slow canonicalization between two fast ones (though the implementation doing the sum canonicalization was left alone and thus still uses a multi-pass procedure). Finally, the new implementation supports canonicalization of tensors with auxiliary indices.
- Some more or less minor changes/refactoring - see individual commits for summaries 

Additional notes:
- I tried adapting the wick implementation's dependence on the graph-representation of a tensor network but failed to get things working together. Thus, I have re-introduced the old graph-representation as a new `WickGraph` class that will be used for Wick. At some point we should try wiring up the Wick impl to the new graph representation, but for now I'm afraid I don't have the time to keep trying.

Fixes #161
Fixes #107


Krzmbrzl:
@evaleev I believe this is ready for review now. Note that this PR has a cross-relation to #180 in that if either of these PRs is merged, the other needs to be updated.

/CC @ak-ustutt

Krzmbrzl:
Rebased against current master and thereby resolved the cross-dependence on the hash PR

Krzmbrzl:
Merged with master (no rebase due to too many conflicts with #184 - didn't feel like doing a rebase with those)

evaleev:
@Krzmbrzl thanks, this is quite a chunk of work!!! I just had a look at this ... do you recall what the issue is with the Wick engine not meshing with the new network graph? At least, what the symptoms were?

Krzmbrzl:
> do you recall what the issue is with the Wick engine not meshing with the new network graph? At least, what the symptoms were?

@evaleev At first I ran into various asserts within the Wick impl due to not having resolved an implicit assumption about the graph representation. Those were about e.g. tensor core vertex ID's being usable as indices into the tensor vector of the network.
Once I fixed the code triggering the assertion I still had the issue of producing wrong results. Iirc for test cases I mostly saw failures where my patched code produced one or two additional terms.
Finally, running the `srcc` example, the performance was observed to be terribly degraded compared to the unpatched version. Furthermore, during the derivation of the doubles or triples residual (I think it was the doubles), RAM consumption simply kept increasing until my system ran out of RAM (32 GB). I believe that I somehow created an infinite loop in the recursive wick impl call tree.

Note that I did find some bugs in my new canonicalization routine after having re-introduced the old graph for the Wick program so perhaps (part of) the issue in the end was a buggy graph representation after all.

----

My biggest problem in all of this was that the wick implementation and the graph representation are tightly coupled, but only implicitly. That means that the wick implementation access what should probably be an implementation detail of the tensor network implementation (the exact graph representation).
If this dependence would instead be through a level of indirection (e.g. member function of a graph object (e.g. `vertex_id_to_tensor_index`)), refactoring would probably be a lot easier as then the parts of the code that need to be adapted is strictly limited to the implementation of these indirection functions.

Krzmbrzl:
@evaleev the new SRCC tests for CSVs are failing. One of them fails due to an unmet assertion that detected external indices are unique, but this unique-ness is based on `Index::LabelCompare` - this was overtaken from the old implementation. However, in the SRCC example 5 I now get two external indices `{a_2^{{i_1}}}` and `{a_2^{{i_2}}}` that only differ in their proto index label.
I would tend to replace the `LabelCompare` with a `FullLabelCompare`, but I'm not sure what the expected behavior is in this case as I have never worked with CSVs. Any insights from your end?

Krzmbrzl:
Superseded by #243

PR #187, opened 2024-03-13
Conner local/runtime_IndexSpace

cmasteran1:
NOT FINISHED. Main goal is to generalize IndexSpace for runtime construction.

changes:
- `Context` contains an `IndexSpaceRegistry`, instead of `IndexSpaceRegistry` being a static member of `IndexSpace`. This gives the user freedom to define set-theoretic relationships on the spaces (union, intersect, order), and removes the need to have fixed definitions for standard spaces.
- To use `SingleProduct` fermionic algebra and operators, the user must additionally define a space which is `vacuum_occupied`. This change naturally leads to more general operator construction (no need to distinguish whether an operator is single-reference or multi-reference. The operator form can be deduced from the IndexSpaceRegistry). This leads to significant removal of redundant code in `mbpt`.


SeQuant/core/space.hpp
```diff
@@ -417,109 +314,446 @@ class IndexSpace {
...
+    base_key_ = other.get_base_key();
+    approximate_size_ = other.get_approximate_size();
     return *this;
   }
 
-  /// @brief constructs an instance of an IndexSpace object
-  IndexSpace(Type type, QuantumNumbers qns) noexcept : attr_(type, qns) {
-    assert(attr_.is_valid());
+
+  const Attr get_attr()const {
```
Krzmbrzl, line 362:
```suggestion
  Attr get_attr()const {
```
`const` on non-reference/pointer return values are usually not desirable: see also https://stackoverflow.com/a/8716406

SeQuant/core/space.hpp
```diff
@@ -417,109 +314,446 @@ class IndexSpace {
...
+
+  // having approximate sizes of spaces can be helpful when optimizing evaluation order.
+  unsigned long get_approximate_size() const {
+    return approximate_size_;
   }
 
  private:
   Attr attr_ = Attr::invalid();
+  std::wstring base_key_;
+  unsigned long approximate_size_;
```
Krzmbrzl, line 386:
```suggestion
  std::size_t approximate_size_;
```
could be seen as more idiomatic: use size_t for anything representing a size

SeQuant/core/space.hpp
```diff
@@ -417,109 +314,446 @@ class IndexSpace {
...
-    assert(attr.is_valid());
+  //explicit IndexSpace(Attr attr) noexcept : attr_(attr) {
+  //  assert(attr.is_valid());
+  //}
+
+
+
+  static std::wstring to_wstring(std::wstring_view key) {
+    return std::wstring(key.begin(), key.end());
   }
```
Krzmbrzl, line 396: What is this used for? It should be completely equivalent to simply writing `std::wstring(key)` - implementing a wrapper function for this doesn't seem to gain much

SeQuant/core/space.hpp
```diff
@@ -417,109 +314,446 @@ class IndexSpace {
...
+};
 
-  static std::wstring_view reduce_key(std::wstring_view key) {
-    const auto underscore_position = key.rfind(L'_');
-    return key.substr(0, underscore_position);
+// An IndexSpaceRegistry is a context specific, modifiable, list of registered indices.
+// It is the task of the USER to ensure that all spaces for their context are included.
+// the result of certain set operations such as intersection and unIon require access to the registry
+//
+class IndexSpaceRegistry{
```
Krzmbrzl, line 404: I would propose putting this into its own header file for better discoverability

SeQuant/core/space.hpp
```diff
@@ -417,109 +314,446 @@ class IndexSpace {
...
-    throw bad_key();
+  //retrieve an IndexSpace from the registry by the label
+  IndexSpace retrieve(std::wstring_view label) const{
+    for(auto it = label_space.begin(); it != label_space.end(); ++it){
+      if(IndexSpace::reduce_key(label) == it->first){
+        return it->second;
+      }
+    }
+    throw std::invalid_argument("The label you are trying to retrieve has not been registered");
+    return label_space.at(L"");
```
Krzmbrzl, line 419: Unreachable code

SeQuant/core/space.hpp
```diff
@@ -417,109 +314,446 @@ class IndexSpace {
...
-    if (found_it != key2attr_.end()) return found_it->second;
-    throw bad_key();
+  //retrieve an IndexSpace from the registry by the label
+  IndexSpace retrieve(std::wstring_view label) const{
+    for(auto it = label_space.begin(); it != label_space.end(); ++it){
+      if(IndexSpace::reduce_key(label) == it->first){
+        return it->second;
+      }
+    }
+    throw std::invalid_argument("The label you are trying to retrieve has not been registered");
```
Krzmbrzl, line 418: It'd be nice if the error message contained the key that wasn't found

SeQuant/core/space.hpp
```diff
@@ -417,109 +314,446 @@ class IndexSpace {
...
+    return label_space.at(L"");
   }
 
-  static std::wstring to_wstring(std::wstring_view key) {
-    return std::wstring(key.begin(), key.end());
+  //add an IndexSpace to this registry. duplicate type bitsets forbidden
+  void add(IndexSpace IS){
+    if(!(find_IndexSpace(IS) == nulltype) ){
```
Krzmbrzl, line 424:
```suggestion
    if(find_IndexSpace(IS) != nulltype){
```
(for easier readability - the leading ! can be easily missed along the parenthesis)

SeQuant/core/space.hpp
```diff
@@ -417,109 +314,446 @@ class IndexSpace {
...
+if(label_space.find(label) == label_space.end()){
+throw std::invalid_argument("label not added to registry");
+}
+density_occuiped_ = label_space[label];
+}
+
+
+ private:
+  std::map<std::wstring ,IndexSpace> label_space;
+  const IndexSpace nulltype= {L"",0,0};
```
Krzmbrzl, line 708:
```suggestion
  static const IndexSpace nulltype= {L"",0,0};
```

SeQuant/core/space.hpp
```diff
@@ -417,109 +314,446 @@ class IndexSpace {
...
+
+  ///TODO use c++20 std::has_single_bit() when we update to this version
+  bool has_single_bit(std::uint32_t bits){
+    return bits&(((bool)(bits&(bits-1)))-1);
```
Krzmbrzl, line 712:
```suggestion
    return bits && !(bits & (bits - 1));
```
conversion from integral types to booleans happens implicit

SeQuant/core/space.hpp
```diff
@@ -417,109 +314,446 @@ class IndexSpace {
...
+  }
+  // find an indexspace from its type. return nullspace if not present.
+  // a bit strange, but prevents an additional map that needs to be maintained
+  const IndexSpace find_IndexSpace(IndexSpace IS)const{
```
Krzmbrzl, line 716:
```suggestion
  IndexSpace find_IndexSpace(const IndexSpace &IS)const{
```
The same suggestion also applies to other functions in the registry class. An alternative would be to return a const _reference_ to the index space to avoid downstream code having to copy the space object if they are only interested in a property of it.

SeQuant/core/space.hpp
```diff
@@ -417,109 +314,446 @@ class IndexSpace {
...
+  // sometimes we wish to check whether if an antribute is in the registry
+const IndexSpace find_IndexSpace(IndexSpace::Attr attr)const{
+  for(auto it = label_space.begin(); it != label_space.end(); it++){
+    if(it->second.attr() == attr){
+      return it->second;
+    }
   }
+  return nulltype;
+}
+  IndexSpace complete_ = {L"",0,0}; // need to define to use generic operator class
```
Krzmbrzl, line 735: Complete here means that this is supposed to be the (implicit) union of all registered index spaces, correct?
If so, a term that I have seen more often in this context is "general" rather than "complete" but that may very well be subjective perception.

SeQuant/core/space.hpp
```diff
@@ -417,109 +314,446 @@ class IndexSpace {
...
+  IndexSpace vacuum_occupied_ = {L"",0,0}; // needed for fermi vacuum wick application
+  IndexSpace density_occuiped_ = {L"",0,0};
+  // both needed to make excitation and de-excitation operators. not neccessarily exclusionary in the case of
+  // multi-reference context.
+  IndexSpace active_particle_space_ = {L"",0,0};
+  IndexSpace active_hole_space_ = {L"",0,0};
```
Krzmbrzl, line 741: Why not use default constructors?
```suggestion
  IndexSpace complete_; // need to define to use generic operator class
  IndexSpace vacuum_occupied_; // needed for fermi vacuum wick application
  IndexSpace density_occupied_;
  // both needed to make excitation and de-excitation operators. not neccessarily exclusionary in the case of
  // multi-reference context.
  IndexSpace active_particle_space_;
  IndexSpace active_hole_space_;
```

SeQuant/core/space.hpp
```diff
@@ -417,109 +314,446 @@ class IndexSpace {
...
+      if(!same_qns){throw std::invalid_argument("asking for the intersection of spaces with incompatible quantum number attributes.");}
+      auto intersection_attr = space1.type().intersection(space2.type());
+      IndexSpace intersection_space = find_IndexSpace({intersection_attr,space1.qns()});
+      // the nullspace is a reasonable return value for intersection
+      if(intersection_space == nulltype && intersection_attr != 0){
+        throw std::invalid_argument("The resulting space is not registered in this context. Add this "
+            "space to the registry with a label to use it.");
+      }
+      else{return intersection_space;}
+    }
```
Krzmbrzl, line 520: After branches that end in a `return` or a `throw`, it is not necessary to write following code in an `else` statement. Omitting the `else` has the advantage of not having to indent that following code by another level which tends to make things more readable if multiple nested indentation levels are at play.

SeQuant/domain/mbpt/convention.cpp
```diff
@@ -62,118 +28,138 @@ auto qndecorate(qns qn, std::wstring_view label) {
...
+  standard_reference_registry.add(complete_unoccupied);
+
+  IndexSpace all(L"p",unoccupied.type().unIon(occupied.type()).to_int32());
+  standard_reference_registry.add(all);
+
+  IndexSpace complete(L"κ",complete_unoccupied.type().unIon(occupied.type()).to_int32());
+  standard_reference_registry.add(complete);
+
+  //only neccessary for some Operator definitions
+  standard_reference_registry.assign_density_occupied(L"m");
```
Krzmbrzl, line 69: Maybe an overload for the `assign` functions would be nice that takes an index space object and then internally resorts to adding the index space and associating its label with the respective group. That would allow to configure such spaces with a single function call instead of two.

SeQuant/core/space.hpp
```diff
@@ -417,109 +314,446 @@ class IndexSpace {
...
+  return nulltype;
+ }
+
+// pure occupied means all states are occupied in a given reference
+bool is_pure_occupied(IndexSpace IS)const{
+ if(IS == nulltype_()){
+    return  false;
+}
+   if(IS.type().to_int32() <= vacuum_occupied().type().to_int32()){
+return true;
```
Krzmbrzl, line 595: This implies numeric ordering of bit patterns which (at least to me) is not obvious. I would have expected bit patterns to only indicate whether spaces overlap or not. If the numeric value is taken into account like it is here, this needs to be documented (rather visibly, I would say).

Wouldn't it be possible to leverage the overlap mechanic for this by having the user define `vacuum_occupied` as the union of all possible sub-spaces that are occupied in the vacuum? In that case this function could check that there is an intersection between the given spaces.

SeQuant/core/space.hpp
```diff
@@ -417,109 +314,446 @@ class IndexSpace {
...
+      }
+      result.push_back(excluded_space);
+    }
+    return result;
+  }
+
+// do two spaces have non_overlapping spaces
+bool has_non_overlapping_spaces(const IndexSpace &space1, const IndexSpace &space2)const{
+if(space1.type().exclusionary_or(space2.type()).to_int32() == 0) {return false;}
+else{ return true;}
```
Krzmbrzl, line 583:
```suggestion
return space1.type().exclusionary_or(space2.type()).to_int32() != 0;
```
No need for any branching when all you want to do is return a bool

SeQuant/core/space.hpp
```diff
@@ -417,109 +314,446 @@ class IndexSpace {
...
+      if(it != label_space.end() && !found){
+        it++;
+      }
+    }
+    if(!found){throw "orginal label not found!";}
+  }
+
+
+  //pass a function which computes a logical bit operation between two IndexSpace.type()
+  const bool valid_bitop( const IndexSpace i1, const IndexSpace i2, const std::function<int32_t(int32_t,int32_t)> op) {
```
Krzmbrzl, line 494:
```suggestion
  bool valid_bitop( const IndexSpace &i1, const IndexSpace &i2, const std::function<int32_t(int32_t,int32_t)> &op) {
```

SeQuant/core/space.hpp
```diff
@@ -417,109 +314,446 @@ class IndexSpace {
...
+      return a.first.type() < b.first.type();
+    });
+    return result;
+  }
+
+  // clear the label_space map essentially clearing the registry
+  void clear_registry(){
+    label_space.clear();
+    IndexSpace nulltype(L"",0,0);
+    this->add(nulltype);
```
Krzmbrzl, line 455:
```suggestion
    this->add(nulltype);
```
Presumably, we want to reuse the class-level `nulltype` instance

Krzmbrzl (commented):
I'm thinking that the user could be freed of the burden of defining all intersection and union types explicitly if they don't mind about the specific labels used by adding a function `IndexSpaceRegistry::createMissingSpaces()` that will automatically check all possible unions and intersections of registered spaces and if no space is registered for this union/intersection, create a new space with an auto-assigned space label (that must not conflict with an already assigned label).
As long as the label-association is logged (e.g. `'a' is the union of 'b'  and 'c'), this should still allow users to figure out what a given space actually means.

SeQuant/core/space.hpp
cmasteran1: This does imply an ordering. I am assuming that all registries which deal with fermions should be ordered such that occupied type orbitals come before unoccupied type orbitals. I will document this better. I assume this is a reasonable assumption for any practical choice of partitioning, is there a case where this might not be true?

Krzmbrzl: > I assume this is a reasonable assumption for any practical choice of partitioning, is there a case where this might not be true?

It might be, but I think we don't really need it. The index space concept would (from my POV) be simpler if it involved only a single concept (bit patterns to describe overlapping spaces) instead of involving a second, unrelated concept (ordering based on the numeric values of those bit patterns).

Krzmbrzl: Additionally, this logic can only ever be applied to a single orbital space group as all other groups would necessarily have higher values where the first group (occupied) would necessarily have lower values and thus by this logic be included in the check.
And if the check was to be extended with a lower bound as to not overlap with lower spaces, this would implicitly end up pre-defining  existing index spaces which would make defining nee ones more involved (as they have to evade falling into one of those implicit assumptions)

SeQuant/core/context.cpp
```diff
@@ -31,15 +31,17 @@ set_scoped_default_context(const Context& ctx) {
...
                  SPBasis spb, std::size_t fdio)
     : vacuum_(vac),
+      idx_space_reg_(std::make_shared<IndexSpaceRegistry>(isr)),
       metric_(m),
       braket_symmetry_(bks),
       spbasis_(spb),
       first_dummy_index_ordinal_(fdio) {}
 
 Vacuum Context::vacuum() const { return vacuum_; }
+std::shared_ptr<IndexSpaceRegistry> Context::index_space_registry() const {return idx_space_reg_;};
```
evaleev, line 44: @cmasteran1 you must install `pre-commit` git hooks and use it throughout. Autoformatting is the only sane way ...

tests/unit/test_parse_expr.cpp
```diff
@@ -80,19 +80,8 @@ ParseErrorMatcher parseErrorMatches(std::size_t offset, std::size_t length,
...
-  minimal_sp.add(iup);
-  IndexSpace idown(L"i", 0b01, mbpt::Spin::beta);
-  minimal_sp.add(idown);
-  IndexSpace anull(L"a", 0b10, mbpt::Spin::any);
-  minimal_sp.add(anull);
-  IndexSpace aup(L"a", {0b10}, mbpt::Spin::alpha);
-  minimal_sp.add(aup);
-  IndexSpace adown(L"a", {0b10}, mbpt::Spin::beta);
-  minimal_sp.add(adown);
+  IndexSpaceRegistry minimal_sp = mbpt::make_min_sr_subspaces();
```
evaleev, line 83: `make_min_sr_subspaces()` already calls `add_fermi_spin()` ...

SeQuant/core/expr.cpp
```diff
@@ -378,7 +378,8 @@ ExprPtr Sum::canonicalize_impl(bool multipass) {
     // ... then reduce terms whose hash values are identical
     auto first_it = begin(summands_);
     auto hash_comparer = [](const auto &first, const auto &second) {
-      return first->hash_value() == second->hash_value();
+      bool first_is_second = first->hash_value() == second->hash_value();
+      return first_is_second;
```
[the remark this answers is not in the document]
evaleev: fixed

SeQuant/core/space.hpp
```diff
@@ -5,17 +5,14 @@
...
-#include <stdexcept>
-#include <string>
-#include <string_view>
-#include <utility>
+#include <cmath>
 
-#include <SeQuant/core/container.hpp>
+#include "attr.hpp"
+#include "container.hpp"
+
```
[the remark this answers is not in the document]
evaleev: addressed via 71bb340a4b51fbe549e211bd2b5c0cedc9854b40

evaleev (approved):
Still will need tweaks to API and cleanup ... but mostly there, so makes sense to merge so that other branches can adjust

PR #197, opened 2024-05-03
find or fetch catch2, use slim headers

evaleev:


Krzmbrzl:
@evaleev this breaks with the find-or-fetch policy. On my system I have Catch2 installed, but I don't have the `catch_test_macros.hpp` headers available. I only have the combined `catch.hpp` header. 

See also https://github.com/catchorg/Catch2/issues/2153

So it seems that we either have to
- Always fetch Catch2 and use that
- Go back to including the main `catch.hpp` header
- Perform explicit checks to see if the `catch_test_macros` exists and if it doesn't fall back to fetching Catch2 after all

PR #203, opened 2024-06-13
Use `IndexSpaceRegistry::approximate_size` in optimization

bimalgaudel:


SeQuant/core/index_space_registry.hpp
```diff
@@ -157,6 +169,20 @@ class IndexSpaceRegistry {
...
+  /// @return pointer to the IndexSpace associated with that key, or nullptr if
+  /// not found
+  IndexSpace* retrieve_ptr(const IndexSpace::Type& type,
+                           const IndexSpace::QuantumNumbers& qns) {
+    auto it = std::find_if(
+        spaces_->begin(), spaces_->end(),
+        [&](const auto& is) { return is.type() == type && is.qns() == qns; });
+    return it != spaces_->end() ? &(*it) : nullptr;
+  }
+
```
ajay-mk, line 185: Can we add one more overload here?
```C++
template <typename S, typename = meta::EnableIfAllBasicStringConvertible<S>>
IndexSpace retrieve(S&& label, const IndexSpace::QuantumNumbers& qns) {}
```
This will make it easier in MPQC to retrieve spaces with spin labels.

bimalgaudel: That will throw when the index space does not exist. [See here.](https://github.com/ValeevGroup/SeQuant/blob/4c1eaad2802bf6a136aae266f41822d708c1cfdd/SeQuant/core/index_space_registry.hpp#L155). The idea is to let the user decide what to do in that scenario, they must test if nullptr is returned. Besides, we need to return `IndexSpace &` not `IndexSpace` for a mutable IndexSpace access even if we do add such overloads. @evaleev any thoughts?

bimalgaudel: If you meant to add `retrieve_ptr` overload with label and quantum numbers overload, sure that aligns well with the added overloads in this PR.

bimalgaudel: @ajay-mk I realize we don't need these overloads. See my recent changes. Just use `retrieve_ptr` with label argument. The label (`wstring_view`) can contain the characters that denote spin-annotation. Let me know if that works.

ajay-mk: Yes, that works.

ajay-mk (approved):

PR #209, opened 2024-06-18
`constexpr` reserved bitset and null bitset.

bimalgaudel:


SeQuant/core/space.hpp
```diff
@@ -252,7 +255,7 @@ class QuantumNumbersAttr {
   /// makes reserved object
   static QuantumNumbersAttr make_reserved() {
     QuantumNumbersAttr result;
-    result.bitset = 0x80000000;
+    result.bitset = TypeAttr::reserved_bitset;
```
evaleev, line 258: clearly `reserved`/`null` bitsets are used more widely than in `TypeAttr`, so move them to `sequant` ns

evaleev: possibly even to
```
namespace sequant {
  namespace bitset {
    using type = int32_t;
    constexpr type reserved = 0x80000000;
    constexpr type null = 0x00000000;
  }
  using bitset_t = bitset::type;
}
```

evaleev (changes requested):

evaleev (approved):

PR #211, opened 2024-06-20
Factor out the gcd of the scalars when two `Product` summands are fused.

bimalgaudel:
Eg.
 - `1/2 AB + 1/4 AC = 1/4 A (2 B + C)`
 - `- 1/2 AB - 1/4 AC = - 1/4 A (2 B + C)`

In the second example, the gcd is factored out, and -1 is factored out afterward as a canonicalization step.

### DIFF
--- a/SeQuant/core/optimize/fusion.cpp
+++ b/SeQuant/core/optimize/fusion.cpp
@@ -49,12 +49,13 @@ ExprPtr Fusion::fuse_left(Product const& lhs, Product const& rhs) {
   auto lsmand_prod = Product{lsmand.begin(), lsmand.end()};
   auto rsmand_prod = Product{rsmand.begin(), rsmand.end()};
 
-  if (lhs.scalar() == rhs.scalar())
-    fac_prod.scale(lhs.scalar());
-  else {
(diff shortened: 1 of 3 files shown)


evaleev (approved):

SeQuant/core/optimize/fusion.cpp
```diff
@@ -102,4 +104,29 @@ ExprPtr Fusion::fuse_right(Product const& lhs, Product const& rhs) {
   return ex<Product>(ExprPtrList{ex<Sum>(ExprPtrList{a, b}), f});
 }
 
+rational Fusion::gcd_rational(rational const& left, rational const& right) {
+  auto&& r1 = left.real();
+  auto&& r2 = right.real();
+  auto&& n1 = numerator(r1);
+  auto&& d1 = denominator(r1);
+  auto&& n2 = numerator(r2);
+  auto&& d2 = denominator(r2);
```
Krzmbrzl, line 113: These should be `const auto &` - forwarding reference are only useful in combination with `std::forward`, which in turn is only useful if data needs to be copied around.

PR #215, opened 2024-07-03
Improve clang format script

Krzmbrzl:


Krzmbrzl:
@evaleev do the scripts (still) work on macOS?

evaleev:
> @evaleev do the scripts (still) work on macOS?

appears to. What was the issue on "your" platform?

evaleev (approved):

Krzmbrzl:
> What was the issue on "your" platform?

Turning paths into relative ones failed because I have `~/Documents/` symlinked to a directory on a different drive. And since `SeQuant` lives at `~/Documents/Git/SeQuant` its path expanded to something like `/other/drive/SeQuant` whereas my home still is `/home/robert`. Thus, the used algorithm for turning paths inside the SeQuant dir to paths relative to the home directory failed.
In the new script I instead make paths relative to the SeQuant root dir itself.

And while I was doing that I found out that we don't have to code the "making a path relative to another" thing ourselfs.

For completeness sake, I am working on Linux (KDE Neon to be precice)

evaleev:
Thanks!

PR #216, opened 2024-07-03
Introduce ResultExpr

Krzmbrzl:
This is a wrapper type that allows to encode the LHS (the result) as well as the LHS (the expression) of an equation. Explicitly tracking the LHS is required in order to access information about which indices are paired together during spin integration (the approach abusing the symmetrizer operators breaks down for results that have mixed index spaces in the bra and/or ket).

In the future, this class may solve as a nucleation point for a framework to store arbitrary metadata about a given expression because it is not nestable (contrary to Expr objects).


SeQuant/core/result_expr.hpp
```diff
@@ -0,0 +1,86 @@
...
+#include <initializer_list>
+#include <optional>
+#include <string>
+
+namespace sequant {
+
+class Tensor;
+class Variable;
+
+class ResultExpr {
```
evaleev, line 19: dox plz

Krzmbrzl:
There will be a follow-up PR that contains this change as well as others necessary for SeQuant to be able to deal with MR theories

PR #218, opened 2024-07-04
Miscellaneous improvements

Krzmbrzl:


Krzmbrzl:
Hm... apparently the tags are no longer cleared. That seems like a bug somewhere in the `Tensor::reset_tags` implementation to me :thinking:

Krzmbrzl:
> That seems like a bug somewhere in the Tensor::reset_tags implementation to me

Nope - the error was on me

PR #220, opened 2024-07-04
Export framework

Krzmbrzl:
This implements (more or less) general code generator capabilities into SeQuant. The idea is to abstract the bulk of the work required during code generation away such that new languages can be added as easily as possible.

----

### TODO
- [x] Verify index order of intermediates is still uniquely determined (sorted)
- [x] Make test cases work
- [x] Get rid of  `insert_blank_lines` function of `Generator` as it breaks the model of only generating semantics
- [x] Allow specifying name of a computed result
- [x] Support multiple expressions in a single export
- [x] Support `ResultExpr` objects
- [x] Migrate ITF exporter to new framework
- [x] Test migrated ITF exporter


Krzmbrzl:
The optimizer still has some quirks (hence it's disabled by default) but the rest seems to be working as intended. Thus, this can be merged as far as I am concerned.

SeQuant/core/utility/atomic.hpp
```diff
@@ -0,0 +1,25 @@
+#ifndef SEQUANT_UTILITY_ATOMIC_HPP
+#define SEQUANT_UTILITY_ATOMIC_HPP
+
+#include <atomic>
+#include <concepts>
+
+namespace sequant {
+
+template <std::integral T>
```
ajay-mk, line 9: Please add docs

SeQuant/core/expressions/tensor.hpp
```diff
@@ -250,12 +251,28 @@ class Tensor : public Expr, public AbstractTensor, public Labeled {
 
   /// @return "core" label of the tensor
   std::wstring_view label() const override { return label_; }
+  void set_label(std::wstring label) {
```
ajay-mk, line 254: Please add docs for the new functions

SeQuant/core/expressions/variable.hpp
```diff
@@ -0,0 +1,67 @@
...
+  template <typename U, typename = std::enable_if_t<!is_variable_v<U>>>
+  explicit Variable(U &&label) : label_(std::forward<U>(label)) {}
+
+  Variable(std::wstring label) : label_(std::move(label)), conjugated_(false) {}
+
+  /// @return variable label
+  /// @warning conjugation does not change it
+  std::wstring_view label() const override;
+
+  void set_label(std::wstring label);
```
ajay-mk, line 32: Better to move `set_label()` to `Labeled` base class?

ajay-mk (commented):
Going through the PR, about half-way through. Here are the first set comments:

- Some parts of the Export code needs docs, for example the `ExportContext` class.
- Ideally, a major refactor like [a838764](https://github.com/ValeevGroup/SeQuant/pull/220/commits/a838764baae4a2433e6525eec11c7c6be01e8fa2) would have been better handled in its own PR. Since the code has been moved into new files, it's now difficult to track what has actually changed in those parts, if anything.

Krzmbrzl:
> Ideally, a major refactor like [a838764](https://github.com/ValeevGroup/SeQuant/pull/220/commits/a838764baae4a2433e6525eec11c7c6be01e8fa2) would have been better handled in its own PR. Since the code has been moved into new files, it's now difficult to track what has actually changed in those parts, if anything.

@ajay-mk you can simply filter out the changes from https://github.com/ValeevGroup/SeQuant/pull/220/commits/a838764baae4a2433e6525eec11c7c6be01e8fa2 in the UI (or rather restrict to the commits before that).

image

In fact, simply by checking the commit message, you should get a pretty good idea about what commit does what and hence which parts of the code are likely affected.

ajay-mk (changes requested):
Here are the remaining comments:
- As mentioned earlier, some parts are still missing documentation.
- It would be helpful to eventually have a basic tutorial on how to use and develop the export framework. This isn’t urgent, but something to consider for the future.
- One concern is the inconsistency in code style. For VRG code, we’ve been following these conventions:
	- `CamelCase` for class and struct names
	- `snake_case` for functions and variables
	- Private members ending with an underscore (`_`)
- It would be great to adopt and enforce a consistent style across the project. There is already some code in master that doesn't follow this style.
cc: @evaleev

Krzmbrzl:
> It would be great to adopt and enforce a consistent style across the project. There is already some code in master that doesn't follow this style.

I agree. Given that this doesn't only affect this PR, I would suggest to do code style unification in a separate PR once this one is merged.

Krzmbrzl:
@CodiumAI-Agent /improve

PR #222, opened 2024-07-08
`Singleton` is thread-safe

evaleev:
- `Singleton::get_instance*` -> `Singleton::instance*`
- to manage lifetime API not returns `shared_ptr` instead of raw handles
- `Singleton::instance()` is safe to call from multiple threads even if `Singleton::set_instance` had not been called

### DIFF
--- a/SeQuant/core/utility/singleton.hpp
+++ b/SeQuant/core/utility/singleton.hpp
@@ -7,6 +7,7 @@
 
 #include <cassert>
 #include <memory>
+#include <mutex>
 #include <stdexcept>
 #include <type_traits>
 
@@ -25,10 +26,12 @@ namespace sequant {
 /// Here's how to use it:
 /// \code
 /// A::set_instance(Args...);  // this creates the first instance of A, if A is not default-constructible, otherwise can skip this
-/// A& the_instance_ref = A::get_instance();      // throws if the instance of A had not been created
-/// A* the_instance_ptr = A::get_instance_ptr();  // returns nullptr if the instance of A had not been created
+/// std::shared_ptr<A> the_instance_ref = A::instance();      // throws if the instance of A had not been created
+/// std::shared_ptr<A> the_instance_ptr = A::instance_ptr();  // returns nullptr if the instance of A had not been created
+/// ... use the_instance_ref or the_instance_ptr ...
 /// // the instance of A will be destroyed with other static-linkage objects
 /// \endcode
+/// Singleton is thread-safe, in that multiple threads can call `A::instance()`, even if `A::set_instance()` has not been called yet.
 // clang-format on
 template <typename Derived>
 class Singleton {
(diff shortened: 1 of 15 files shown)


Krzmbrzl:
I don't think the thread-safety is fulfilled. As I see it, if two threads simultaneously call `Singleton::instance` whilst the singleton hasn't been constructed yet, we get two independent calls to `Singleton::set_instance()`. These two calls are synchronized via the mutex such that they will execute sequentially. In both of these calls, a new singleton instances is created and assigned to the instance accessor. That means that the second call will overwrite the first one.
This violates the core concept of what a singleton is supposed to provide (only a single instance can ever be created).

evaleev:
> I don't think the thread-safety is fulfilled. As I see it, if two threads simultaneously call `Singleton::instance` whilst the singleton hasn't been constructed yet, we get two independent calls to `Singleton::set_instance()`. These two calls are synchronized via the mutex such that they will execute sequentially. In both of these calls, a new singleton instances is created and assigned to the instance accessor. That means that the second call will overwrite the first one. This violates the core concept of what a singleton is supposed to provide (only a single instance can ever be created).

I agree that it is possible to have `Singleton::instance()` return different objects. E.g., calling Singleton::instance() from multiple threads without having called `Singleton::set_instance()`. But mutable Singleton concept is not well defined in presence of threads. I.e., if `Singleton::set_instance()` can be called multiple times (e.g. to reset Singleton instance) I'm not sure if there is natural and safe semantics that does not rely on `set_instance` being called from a single thread.

I'm happy to ensure that a single default instance. But returning shared_ptr means that multiple copies can be around at once, or we can have a deadlock (i.e. waiting for all refs to instance to disappear before creating a new one may cause a deadlock), or we revert back to bare handles and UB ...

SeQuant/core/utility/singleton.hpp
```diff
@@ -75,21 +78,39 @@ class Singleton {
...
   /// @param args a parameter pack
+  /// @return shared_ptr to the newly-created instance
   template <typename... Args>
-  static void set_instance(Args&&... args) {
-    assert(instance_accessor() == nullptr);
+  static std::shared_ptr<Derived> set_instance(Args&&... args) {
+    std::scoped_lock lock(instance_mutex());
     instance_accessor() = std::move(
-        std::unique_ptr<Derived>(new Derived(std::forward<Args>(args)...)));
+        std::shared_ptr<Derived>(new Derived(std::forward<Args>(args)...)));
```
Krzmbrzl, line 86: This should probably be
```
instance_accessor() = std::make_shared(std::forward<Args>(args)...);
```

evaleev: no, this is correct to allow using private ctors of Derived.

Krzmbrzl:
Now that I think of it - what even is the semantics we would like to achieve here? Presumably, using the `set_instance` function is only valid while no other thread is running a processing that has started with the old singleton instance. Because even if some part of that other thread has currently a shard pointer and would therefore continue to use the same object for a bit, further down the processing line a different function is likely to do a fresh `instance()` retrieval which would now yield a different instance leading to inconsistencies within the processing.

If my above argument is valid, then there is no need to support one part of the code holding on to an old copy of the singleton while another part has already updated the instance. That would mean that we don't gain anything by using `shared_ptr`s.
In order to prevent dangling pointers, we could instead allocate the instance in a fixed place (either as a direct member variable of `Singleton<Type>` or at a fixed memory address) and once someone calls `set_instance`, we don't allocate a new object but instead overwrite the already existing object in-place. That keeps all existing pointers valid as the memory location hasn't changed.

Nonetheless, having a non-const object shared between multiple threads is inherently unsafe unless that object itself implements means of synchronization. I don't think that we can solve this issue at the `Singleton` level :thinking:

evaleev:
The most common use case is a singleton that is initialized and mutated by a single thread. The only complication is handling the auto-initialization. I think we should ensure that `set_instance` is only called once (i.e. Singleton's instance is created only once). After that returning bare handles should be OK. BUT, since destruction of statics happens in arbitrary order I think using shared_ptr still may make sense to avoid dangling refs/ptrs.

Krzmbrzl:
> BUT, since destruction of statics happens in arbitrary order I think using shared_ptr still may make sense to avoid dangling refs/ptrs.

This would only affect code that itself uses the singleton as part of a static, no? Is there an actual use-case for which this is a valid thing to do?

I think making `instance()` return a pointer makes the API worse as now there is (at least semantically) the possibility of that pointer being `nullptr`. I much prefer to use references for stuff that can't be null as here there is an actual enforcement of this claim and every user can easily see it (whereas for pointers the best you can do is a promise in the documentation of some function).

evaleev:
> This would only affect code that itself uses the singleton as part of a static, no? Is there an actual use-case for which this is a valid thing to do?

At the moment the only example we have is `Logger`, and using it should be possible anywhere (though we do not use it from other static lifetime objects). I guess even such use would work since calling `Logger::instance()` would cause construction of default `Logger`, but that's not what is intended.

Statics are a special (but unavoidable, in header-only context) case, so we can just say "don't do it".

I made https://github.com/ValeevGroup/SeQuant/issues/226 and https://github.com/ValeevGroup/SeQuant/pull/225

PR #224, opened 2024-07-17
subspaces constructed with ignore spin on lead to faulty index compar…

cmasteran1:
Whenever possible please do not ignore spin when constructing subspaces. some functions (like spintrace) construct spinfree indices with Spin::Any QNS. This lead to a case where two identical indices were not identified as such because they had different QNS.


Krzmbrzl:
Thanks for looking into this! :+1:

What is the intended use case for the ignore_spin option?

I'm wondering whether this fixes the underlying issue or only the symptom 🤔

cmasteran1:
As I understand it, the option is only needed to ensure certain unit test canonicalization stays consistent. Considering the option is only present for generating legacy subspaces, I don't think it will be an issue going forward.

tests/unit/test_spin.cpp
```diff
@@ -730,7 +730,7 @@ SECTION("Closed-shell spintrace CCD") {
 
 SECTION("Closed-shell spintrace CCSD") {
   auto ctx_resetter = set_scoped_default_context(
-      Context(sequant::mbpt::make_legacy_spaces(/* ignore_spin = */ true),
+      Context(sequant::mbpt::make_legacy_spaces(/* ignore_spin = */ false),
```
Krzmbrzl, line 733:
```suggestion
      Context(sequant::mbpt::make_legacy_spaces(),
```
The default is `false`

Krzmbrzl:
Okay, I see. I did search the code base for where this is currently used and only found
https://github.com/ValeevGroup/SeQuant/blob/1095dda5c59cbb4aa7793435f02427cc3696622a/tests/unit/test_spin.cpp#L731-L734
which is needed in order for
https://github.com/ValeevGroup/SeQuant/blob/1095dda5c59cbb4aa7793435f02427cc3696622a/tests/unit/test_spin.cpp#L749-L774
to pass

and
https://github.com/ValeevGroup/SeQuant/blob/1095dda5c59cbb4aa7793435f02427cc3696622a/tests/unit/test_spin.cpp#L566-L569
which is needed to make
https://github.com/ValeevGroup/SeQuant/blob/1095dda5c59cbb4aa7793435f02427cc3696622a/tests/unit/test_spin.cpp#L585-L601
pass.

---

To me it is somewhat worrysome that some of the test cases for seemingly foundational functionality only pass with this particular hack :eyes: 

@evaleev do you know why this hack is needed?

tests/unit/test_spin.cpp
```diff
@@ -754,14 +761,13 @@ SECTION("Closed-shell spintrace CCSD") {
...
         L"{{{2}}{g^{{i_1}{a_2}}_{{a_1}{i_2}}}{t^{{i_2}}_{{a_2}}}}\\bigr) }");
-
     container::map<Index, Index> idxmap = {{Index{L"i_1"}, Index{L"i_2"}},
                                            {Index{L"i_2"}, Index{L"i_1"}}};
     auto transformed_result = transform_expr(result, idxmap);
     REQUIRE(
         to_latex(transformed_result) ==
-        L"{ \\bigl( - {{g^{{a_2}{i_1}}_{{a_1}{i_2}}}{t^{{i_2}}_{{a_2}}}} + "
-        L"{{{2}}{g^{{i_1}{a_2}}_{{a_1}{i_2}}}{t^{{i_2}}_{{a_2}}}}\\bigr) }");
+        L"{ \\bigl( - {{g^{{a_2}{i_2}}_{{a_1}{i_1}}}{t^{{i_1}}_{{a_2}}}} + "
```
evaleev, line 769: N.B. this reverts https://github.com/ValeevGroup/SeQuant/commit/fa7ccb9abb9b0e1ec6b9a25416474d9e2ed05bce

evaleev (approved):

evaleev:
> @evaleev do you know why this hack is needed?

I added a note (`WARNING`) explaining the need for this ... potentially easier solution is to change the reference results but it feels good to have a test that uses different spin attribute representation ...

PR #225, opened 2024-07-18
Singleton::instance() returns ref

evaleev:
https://github.com/ValeevGroup/SeQuant/pull/222#issuecomment-2218120202

### DIFF
--- a/SeQuant/core/utility/singleton.hpp
+++ b/SeQuant/core/utility/singleton.hpp
@@ -30,9 +30,9 @@ namespace sequant {
 /// // optional: create the instance of A
 /// A::set_instance(Args...);  // this call-once method creates the instance of A; if A is default-constructible, can skip this as the instance will be created by first call to instance()
 /// // access the instance of A
-/// std::shared_ptr<A> the_instance_ptr = A::instance();      // throws if A is not default-constructible or A::set_instance() had not been called to create the instance
+/// A& the_instance = A::instance();      // throws if A is not default-constructible or A::set_instance() had not been called to create the instance
 /// // ... or ...
-/// std::shared_ptr<A> the_instance_ptr = A::instance_ptr();  // alternative to A::instance(), instead of throwing, returns nullptr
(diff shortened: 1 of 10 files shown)


SeQuant/core/utility/singleton.hpp
```diff
@@ -96,16 +96,16 @@ class Singleton {
           "sequant::Singleton::set_instance: instance has already been "
           "constructed");
     instance_accessor() = std::move(
-        std::shared_ptr<Derived>(new Derived(std::forward<Args>(args)...)));
-    return instance_accessor();
+        std::unique_ptr<Derived>(new Derived(std::forward<Args>(args)...)));
```
Krzmbrzl, line 99: `move` should not be needed here.
Also, we probably should make use of `std::make_unique`

SeQuant/core/utility/singleton.hpp
```diff
@@ -117,12 +117,12 @@ class Singleton {
  private:
   /// Constructs a default-constructed instance. If called from multiple threads
   /// only the first call will construct the instance.
-  static std::shared_ptr<Derived> set_default_instance() {
+  static Derived& set_default_instance() {
     std::scoped_lock lock(instance_mutex());
     if (!instance_accessor()) {
-      instance_accessor() = std::move(std::shared_ptr<Derived>(new Derived));
+      instance_accessor() = std::move(std::unique_ptr<Derived>(new Derived));
```
Krzmbrzl, line 123: move should not be needed here.
Also, we probably should make use of std::make_unique

Krzmbrzl (approved):

SeQuant/core/utility/singleton.hpp
evaleev: can't use `std::make_unique` (see note above re: private ctor) ... agree that `std::move` is not needed but is there a penalty for its use here? as a visual annotation of where moves occur it beats a comment.

PR #204, opened 2024-06-13
Used for successful PNO-CCSD calculations

bimalgaudel:

PR #233, opened 2024-08-29
Merge #204 into master

bimalgaudel:
#204 is there as a time-stamp or a checkpoint to refer to a valid CSV-CCSD evaluation.

We are not going to delete that soon.

Instead, we can use this branch to merge into the master and delete.


evaleev:
@bimalgaudel PRs are not for serving as a code snapshot at particular state (and in fact PRs are purely GitHub artifacts) Particular Git commits is all you need, or you can use Git tags to have a human-readable name. I'd prefer not to keep a bunch of branches just for the purpose of referring to a particular code revision, and certainly there should be no PRs open for that purpose.

please update #204 as requested and we'll merge it. Tag the state you want, or just make sure the MPQC branch documenting the functional CSV-CC points to the correct commit.

PR #242, opened 2024-11-12
Generalize printing of antisymm Tensors

### LINKED ISSUE #241: Printing of Antisymmetric `Tensors` should be more general
Core functions like this should not depend on labels or conventions. 
https://github.com/ValeevGroup/SeQuant/blob/a9206a58c6d2102a80f67a363c9da046bd4a1729/SeQuant/core/tensor.hpp#L220-L225

ajay-mk:
Fixes #241

### DIFF
--- a/SeQuant/core/tensor.hpp
+++ b/SeQuant/core/tensor.hpp
@@ -218,12 +218,11 @@ class Tensor : public Expr, public AbstractTensor, public Labeled {
   }
 
   std::wstring to_latex() const override {
-    std::wstring result;
-    std::vector<std::wstring> labels = {L"g", L"t", L"λ", L"t¹", L"λ¹"};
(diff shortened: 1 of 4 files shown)


ajay-mk:
Copying @bimalgaudel's comment:
In my opinion:

- If Tensor::symmetry() is Symmetry::antisymm, we should add a bar. No matter the number of particles (bra_rank()). Rationale is, we are simply printing the tensor's context.
- If a Tensor::symmetry() has Symmetry::symm we should probably add a hat, following the OpMin paper convention.
- If we want to be precise, the symmetry properties do not actually belong to the tensor label but to some index groups of the bra and ket. This will be particularly important for non-particle conserving quantities.

bimalgaudel (approved):

PR #249, opened 2025-02-02
Bugfix in `evaluate()`

ajay-mk:
Fixes the issue reported in https://github.com/ValeevGroup/mpqc4/pull/521. Removes the extra `add_inplace()` call in `sequant::evaluate()`


evaleev (approved):
Thanks, @ajay-mk

PR #251, opened 2025-02-04
Make TensorBlockIndexComparer compare tensor blocks (slots) again

Krzmbrzl:
This reverts commit ba832da3054fc181fee88d7c45f627a5ea9f1ae7.

The reason being that Index::operator< is NOT the right way to compare indices in this context. If we also account for index labels (instead of only their space), we are comparing full indexed tensor elements rather than tensor blocks (slots) as this class is supposed to do.

In order for test cases to work with this re-introduced change, this commit also reverts
"amend tests for ba832da3054fc181fee88d7c45f627a5ea9f1ae7" (commit 4a9543851105e2bae0939495986351fe74930d70)

Finally, a hand full of additional test cases had to be adapted to the changed canonicalization routine.

### DIFF
--- a/SeQuant/core/tensor_canonicalizer.cpp
+++ b/SeQuant/core/tensor_canonicalizer.cpp
@@ -49,20 +49,80 @@ struct TensorBlockIndexComparer {
       static_assert(std::is_same_v<std::decay_t<decltype(rhs_second)>, Index>,
                     "TensorBlockIndexComparer can only work with indices");
 
-      int res = lhs_first < rhs_first ? -1 : (rhs_first < lhs_first ? 1 : 0);
+      // First compare only index spaces of equivalent pairs
+      int res = compare_spaces(lhs_first, rhs_first);
       if (res != 0) {
         return res;
       }
 
-      res = lhs_second < rhs_second ? -1 : (rhs_second < lhs_second ? 1 : 0);
+      res = compare_spaces(lhs_second, rhs_second);
+      if (res != 0) {
+        return res;
+      }
+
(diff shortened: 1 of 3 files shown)


evaleev:
@Krzmbrzl I agree in principle that we need to distinguish indices from slots (everything in Index but the label) ... do we need to thus always use slots for tensor canonicalization? This PR makes tensor "blocks" (columns?) compare via slots (i.e. labels are ignored), but tensor indices are still compared as indices (with labels taken into account) ...

Krzmbrzl:
I think we need both in order to establish a total order of tensors. Or rather we need to also compare indices for the total ordering and we need to first try and compare only slots if we want consistent results in terms of slot ordering in the canonical form.

So what this PR does is to ensure to first try and establish a total order on slots alone. If this fails, it makes use of the index labels to break ties in the slot ordering. This is needed in case of tensors that are only particle-symmetric and have two indices of the same slot in the bra (which our algorithm compares first): $t^{i_3,a_1}_{i_1, i_2}$
If we compare full indices from the beginning then relabelling  `i_1` to `i_2`  and vice versa would make the canonicalizer swap the columns leading to different slot ordering in the ket indices.

By only taking the slots into account first, we allow the algorithm to also compare the ket indices which yield a total ordering (in this case) based on slots alone (which doesn't change upon relabelling of bra indices).
In the context of TN canonicalization this is especially important because we canonicalize every tensor (block) before relabelling indices (hence the prior index labels don't have that much of a meaning).

evaleev (approved):

PR #257, opened 2025-02-10
Robuster common sub-expression elimination

bimalgaudel:
Use tensor-network canonicalization for common sub-expression identification.


SeQuant/core/eval_expr.cpp
```diff
@@ -459,4 +410,183 @@ ExprPtr make_imed(EvalExpr const& left, EvalExpr const& right,
...
 }  // namespace
 
+///
+/// \brief Sorts the input range and calls hash_range.
+///
+template <typename Rng>
+size_t canon_hash(Rng const& rng) {
+  auto vec = rng | ranges::to_vector;
+  ranges::sort(vec);
+  return hash::hash_range(vec.begin(), vec.end());
```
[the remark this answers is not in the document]
bimalgaudel: This is supposed to be that function. `Rng` could probably be required to be a range of `size_t`.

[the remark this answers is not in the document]
Krzmbrzl: If this is intended to be a commutative hashing function, it might be better to implement it in the `hash.hpp` file so it sits near the other hash functions.

In that case however, it might become even more important to avoid the explicit sorting step. Sorting scales `O(N log(N))` (at best) which can be a lot more expensive than `O(N)` (which you get when simply using a commutative hash combine function to combine the hash for all elements in the range) for big `N`.

SeQuant/core/eval_expr.cpp
```diff
@@ -77,64 +59,67 @@ std::string to_label_annotation(const Index& idx) {
...
-      id_{++global_id_},
       expr_{make_imed(left, right, op)} {
   result_type_ = expr_->is<Tensor>() ? ResultType::Tensor : ResultType::Scalar;
-  tot_ = expr_->is<Tensor>() && is_tot(expr_->as<Tensor>());
+  if (result_type() == ResultType::Tensor) {
+    auto tn = TensorNetwork(expr_);
+    auto canon =
+        tn.canonicalize_slots(TensorCanonicalizer::cardinal_tensor_labels());
+    canon_indices_ = canon.get_indices() | ranges::to<index_vector>;
+  }
```
evaleev, line 122: @bimalgaudel does this need to adjust for the phase also?

bimalgaudel: Sure. But this function is deprecated and will be removed soon.

tests/unit/test_eval_node.cpp
```diff
@@ -253,44 +249,45 @@ TEST_CASE("TEST EVAL_NODE", "[EvalNode]") {
...
-
-    auto const p7 = parse_expr(L"I{i1;a1} * I{i2;a2}", Symmetry::nonsymm);
-    auto const np7 = eval_node(p7);
-    REQUIRE(asy_cost(np7) == AsyCost{{1, 2}, 2, 2});  // 1/2 * O^2V^2
-
-    auto const p8 =
-        parse_expr(L"I{i1,i2;a3,a4} * I{a3,a4;a1,a2}", Symmetry::nonsymm);
-    auto const np8 = eval_node(p8);
-    REQUIRE(asy_cost(np8) == AsyCost{2, 2, 4});  // 2 * O^2V^4
+    // todo
```
evaleev, line 252: cleaner way to disable if by wrapping inside `#if 0 ... #endif`

tests/unit/test_eval_expr.cpp
```diff
@@ -258,23 +259,24 @@ TEST_CASE("TEST_EVAL_EXPR", "[EvalExpr]") {
     const auto t5 = tensor(Symmetry::nonsymm);
     const auto t6 = tensor(Symmetry::nonsymm);
 
+    // todo:
```
evaleev, line 262: cleaner way to disable if by wrapping inside `#if 0 ... #endif`

SeQuant/core/eval_expr.cpp
```diff
@@ -459,4 +385,189 @@ ExprPtr make_imed(EvalExpr const& left, EvalExpr const& right,
...
+      auto canon =
+          tn.canonicalize_slots(TensorCanonicalizer::cardinal_tensor_labels());
+      hash::combine(h, canon.hash_value());
+      bool const scalar_result = ranges::empty(canon.get_indices());
+      if (scalar_result) {
+        return {EvalOp::Prod,            //
+                ResultType::Scalar,      //
+                dummy::make_variable(),  //
+                {},                      //
+                1,                       //
```
evaleev, line 533: presumably need `phase` here also ...

evaleev (changes requested):

evaleev (approved):

PR #266, opened 2025-03-12
Documentation Revamp

ajay-mk:
## Changelist
- All CMake configuration related to documentation now lives in https://github.com/ValeevGroup/SeQuant/blob/ajay/docs/revamp/doc/CMakeLists.txt
- Organized files into folders for user-facing and internal documentation
- Better handling of Python dependencies
- Added notes on building and contributing to documentation. Includes information about the workflow and some basic rst examples
- Added placeholder files for advanced topics. 
- Tried to add `graphviz` support, not fully functional yet. 

See https://ajay-mk.github.io/SeQuant/index.html for a live version. Please take a look, we can move things around if needed.

I think it would be easier if people volunteer to write certain parts of the docs. See https://ajay-mk.github.io/SeQuant/internal/docs_advanced.html for some details to get started.  

#### Further improvements we could do
- We use [Exhale](https://github.com/svenevs/exhale) to generate rst files from Doxygen XML files. This library is not well maintained anymore. We could switch to `breathe-apidoc` which offers similar (but not full) functionality.


doc/CMakeLists.txt
```diff
@@ -1,2 +1,122 @@
-add_subdirectory(doxygen)
-add_subdirectory(sphinx)
+# Main documentation targets
+add_custom_target_subproject(doc sequant)
+
+# Find Python for Sphinx dependencies
+find_package(Python 3.8 COMPONENTS Interpreter REQUIRED)
```
evaleev, line 5: I though you mentioned 3.9 as the minimal version?

doc/CMakeLists.txt
```diff
@@ -1,2 +1,122 @@
...
+    # Link Doxygen to main documentation target
+    add_dependencies(sequant-doc doxygen-sequant)
+    set_property(GLOBAL PROPERTY SEQUANT_HAS_DOXYGEN TRUE)
+else ()
+    set_property(GLOBAL PROPERTY SEQUANT_HAS_DOXYGEN FALSE)
+    message(STATUS "Doxygen not found - Skipping API documentation")
+endif ()
+
+# Find Sphinx for documentation generation
+find_package(Sphinx)
```
evaleev, line 43: no point in discovering Sphinx if Doxygen was not found

doc/CMakeLists.txt
```diff
@@ -1,2 +1,122 @@
-add_subdirectory(doxygen)
-add_subdirectory(sphinx)
+# Main documentation targets
+add_custom_target_subproject(doc sequant)
```
evaleev, line 2:
```suggestion
add_custom_target_subproject(sequant doc)
```
see https://github.com/ValeevGroup/SeQuant/blob/master/cmake/modules/AddCustomTargetSubproject.cmake#L17

doc/CMakeLists.txt
```diff
@@ -1,2 +1,122 @@
...
+    add_custom_target_subproject(sequant doxygen
+            COMMAND ${DOXYGEN_EXECUTABLE} ${DOXYGEN_OUTPUT_DIR}/Doxyfile
+            WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+            COMMENT "Generating API documentation with Doxygen"
+            VERBATIM USES_TERMINAL
+            SOURCES ${DOXYGEN_OUTPUT_DIR}/Doxyfile
+    )
+
+    # Link Doxygen to main documentation target
+    add_dependencies(sequant-doc doxygen-sequant)
```
evaleev, line 35:
```suggestion
    add_dependencies(doc-sequant doxygen-sequant)
```

doc/CMakeLists.txt
```diff
@@ -1,2 +1,122 @@
...
+    endif ()
+
+    # Add Sphinx to main documentation target
+    add_dependencies(sequant-doc sphinx-sequant)
+else ()
+    message(STATUS "Sphinx not found - Skipping documentation generation")
+endif ()
+
+# Alias for HTML generation
+add_custom_target_subproject(sequant html DEPENDS sequant-doc)
```
evaleev, line 122:
```suggestion
add_custom_target_subproject(sequant html DEPENDS doc-sequant)
```

doc/CMakeLists.txt
```diff
@@ -1,2 +1,122 @@
...
+            VERBATIM USES_TERMINAL
+    )
+
+    # Make Sphinx depend on Doxygen if available
+    if (HAS_DOXYGEN)
+        add_dependencies(sphinx-sequant doxygen-sequant)
+    endif ()
+
+    # Add Sphinx to main documentation target
+    add_dependencies(sequant-doc sphinx-sequant)
```
evaleev, line 116:
```suggestion
    add_dependencies(doc-sequant sphinx-sequant)
```

doc/CMakeLists.txt
```diff
@@ -1,2 +1,122 @@
...
+else ()
+    message(STATUS "Sphinx not found - Skipping documentation generation")
+endif ()
+
+# Alias for HTML generation
```
evaleev, line 121:
```suggestion
# `html` is a standard GNU  target: https://www.gnu.org/prep/standards/html_node/Standard-Targets.html#Standard-Targets
```

doc/sphinx/user/docs.rst
```diff
@@ -0,0 +1,24 @@
...
+
+This will install the required packages, including Sphinx, Breathe, Exhale, and others.
+Now you can build the documentation using CMake. The documentation will be generated in the ``doc/html`` directory. You can open the ``index.html`` file in your web browser to view the documentation.
+
+.. code-block:: bash
+
+    mkdir build
+    cd build
+    cmake -DSEQUANT_BUILD_DOCS=ON ..
+    cmake --build . --target sequant-doc
```
evaleev, line 22:
```suggestion
    cmake --build . --target doc-sequant
```

evaleev (changes requested):

doc/CMakeLists.txt
Krzmbrzl: Why? Imo it should be possible to build the user-facing documentation (that is, everything except the immediate in-source doc strings) anyway :thinking:

doc/CMakeLists.txt
```diff
@@ -1,2 +1,122 @@
-add_subdirectory(doxygen)
-add_subdirectory(sphinx)
+# Main documentation targets
+add_custom_target_subproject(doc sequant)
+
+# Find Python for Sphinx dependencies
+find_package(Python 3.8 COMPONENTS Interpreter REQUIRED)
+
+# Find Doxygen for API documentation
+find_package(Doxygen 1.9.3)
```
Krzmbrzl, line 8: Do we really need v1.9.3 or higher? That is, 1.9.2 doesn't work?
_If_ Doxygen was following semantic versioning, I would expect any 1.9.x to work - hence the question

doc/CMakeLists.txt
```diff
@@ -1,2 +1,122 @@
...
+    file(MAKE_DIRECTORY ${DOXYGEN_OUTPUT_DIR})
+
+    # Configure Doxyfile
+    configure_file(
+            ${CMAKE_CURRENT_SOURCE_DIR}/doxygen/Doxyfile.in
+            ${DOXYGEN_OUTPUT_DIR}/Doxyfile @ONLY IMMEDIATE
+    )
+
+    # Set DOT path if available
+    if (DOXYGEN_DOT_EXECUTABLE)
```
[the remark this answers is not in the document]
Krzmbrzl: @ajay-mk just to be clear: there are multiple other locations in which you use the *_EXECUTABLE variables that have been deprecated. You are probably intended to use a generator expression to get the target's executable instead of those variables :thinking:

doc/CMakeLists.txt
ajay-mk: Yes, Breathe requires Python 3.9+ (https://github.com/breathe-doc/breathe#requirements)

doc/CMakeLists.txt
```diff
@@ -1,2 +1,122 @@
...
+        get_filename_component(DOXYGEN_DOT_DIRECTORY ${DOXYGEN_DOT_EXECUTABLE} DIRECTORY)
+        file(TO_NATIVE_PATH ${DOXYGEN_DOT_DIRECTORY} DOXYGEN_DOT_PATH)
+    endif ()
+
+    # Doxygen build target
+    add_custom_target_subproject(sequant doxygen
+            COMMAND ${DOXYGEN_EXECUTABLE} ${DOXYGEN_OUTPUT_DIR}/Doxyfile
```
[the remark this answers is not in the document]
ajay-mk: Updated, now uses `Doxygen::doxygen`

doc/CMakeLists.txt
```diff
@@ -1,2 +1,122 @@
...
+            list(APPEND SPHINX_PYTHON_MODULES ${module_clean})
+        endif ()
+    endforeach ()
+
+    if (SPHINX_MISSING_MODULES)
+        message(FATAL_ERROR
+                "The following Python modules required for Sphinx documentation are missing:\n"
+                "  ${SPHINX_MISSING_MODULES}\n"
+                "Run the following command to install them:\n"
+                "  '${Python_EXECUTABLE} -m pip install -r ${CMAKE_CURRENT_SOURCE_DIR}/requirements.txt'\n"
```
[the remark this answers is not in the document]
ajay-mk: There is a `find_package(Python 3.9 COMPONENTS Interpreter REQUIRED)` at the top of the file. That should avoid the issue right?

doc/CMakeLists.txt
```diff
@@ -1,2 +1,122 @@
...
+    foreach (dir ${DIRS_TO_COPY})
+        if (IS_DIRECTORY ${SPHINX_SOURCE_DIR}/${dir})
+            file(COPY ${SPHINX_SOURCE_DIR}/${dir} DESTINATION ${SPHINX_BUILD_DIR})
+        endif ()
+    endforeach ()
+
+    # Handle legacy source directory
+    if (EXISTS ${SPHINX_SOURCE_DIR}/source)
+        file(COPY ${SPHINX_SOURCE_DIR}/source DESTINATION ${SPHINX_BUILD_DIR})
+    endif ()
```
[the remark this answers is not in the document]
ajay-mk: Yes, this was supposed to be removed.

doc/CMakeLists.txt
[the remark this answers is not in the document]
Krzmbrzl: Ah yeah - I'm blind🤦‍♂️

doc/CMakeLists.txt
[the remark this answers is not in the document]
ajay-mk:
```cmake
    if(TARGET DOXYGEN_HAVE_DOT)
        get_target_property(DOT_EXECUTABLE Doxygen::dot IMPORTED_LOCATION)
        get_filename_component(DOXYGEN_DOT_PATH "${DOT_EXECUTABLE}" DIRECTORY)
        file(TO_NATIVE_PATH "${DOXYGEN_DOT_PATH}" DOXYGEN_DOT_PATH)
    endif()
```
Does that look correct?

doc/CMakeLists.txt
ajay-mk: After doing some digging, I think with the current setup we always need Doxygen. Since we have Exhale directives in the `conf.py`, it looks for `xml` output from Doxygen. It might be possible to avoid this if switch to `breathe-apidoc`, I will try to get that working later. For now changing to:
```cmake
if (SPHINX_FOUND AND DOXYGEN_FOUND)
...
else ()
    message(STATUS "Sphinx not found - Skipping Sphinx documentation generation")
endif ()
```
So the behavior is:

If only Doxygen is found, `html-sequant` will only build Doxygen `xml` and `html`, if Sphinx is found, Sphinx `html` is also generated. If Doxygen is not available, no documentation will be generated.

doc/CMakeLists.txt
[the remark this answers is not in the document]
Krzmbrzl: What is the `DOXYGEN_DOT_PATH` variable used for? I don't see it being used anyswhere :eyes: 

And for the cases where you actually want the path to the executable, the [docs for add_custom_target](https://cmake.org/cmake/help/latest/command/add_custom_target.html) states that for the `COMMAND` parameter
> If COMMAND specifies an executable target name (created by the `add_executable()`), it will automatically be replaced by the location of the executable

so that's what I would try (as I am kinda expecting this to work with imported targets as well as otherwise the switch of FindDoxygen to promote use of the targets rather than the variables doesn't seem to make sense)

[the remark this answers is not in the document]
ajay-mk: Turns out we don't really need the `DOXYGEN_DOT_PATH`. If `dot` is found, we just need to set `HAVE_DOT = YES` in the Doxyfile. So the following should work:
```cmake
    if(DOXYGEN_HAVE_DOT)
        set(DOXYGEN_DOT_FOUND YES)
    endif()
```
In Doxyfle.in
```
HAVE_DOT               = @DOXYGEN_DOT_FOUND@
```
I guess it could be just `HAVE_DOT = @DOXYGEN_HAVE_DOT@`, but good to have a variable we can set to NO if we want to force disable `dot` output.

PR #267, opened 2025-03-12
MR support

Krzmbrzl:
This PR makes SeQuant post-processing ready for being able to deal with multireference theories. In particular, an implicit representation of the result (via e.g. a symmetrization operator) becomes impossible due to the existence of results (residuals) that have indices of different index space in the bra and/or ket group, which must not be symmetrized. The same issue is then also relevant for spintracing, which has to be able to deal with this situation - especially the fact that spintracing a single expression can now lead to producing multiple (independent!) results.

Additionally, this PR introduce what I have called "external interface" which allows to define input equations in SeQuant-parseable form along with a JSON file detailing how the inputs shall be processed. At the moment, this interface is specifically geared towards generating ITF code in the end, but generalizations should be possible.

----

TODO
- [ ] Support spintracing of `ResultExpr` objects for other spintracing variants as well
- [ ] Move MR biorthogonalization logic into SeQuant (instead of living as custom code in the interface)
- [ ] Add MR test cases

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -235,6 +235,8 @@ set(SeQuant_src
         SeQuant/core/parse.hpp
         SeQuant/core/ranges.hpp
         SeQuant/core/rational.hpp
+        SeQuant/core/result_expr.cpp
+        SeQuant/core/result_expr.hpp
         SeQuant/core/runtime.cpp
         SeQuant/core/runtime.hpp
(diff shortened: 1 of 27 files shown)


Krzmbrzl:
Note: The failing test case is exclusively due to lack of idempotency of the tensor network canonicalization. With `TensorNetworkV2`, all tests pass.

PR #268, opened 2025-03-12
Switch over to using TensorNetworkV2

Krzmbrzl:


Krzmbrzl:
The Wick integration needs to be revised but otherwise, all tests are passing. So from my POV, this is ready to merge

PR #271, opened 2025-03-14
More doc improvements

Krzmbrzl:
- It is now possible to only build the Sphinx docs, without generating Doxygen docs. This allows fast prototyping of Sphinx docs without having to wait forever for all the autogenerated stuff to be processed.
- The layout inside `doc` has been adjusted in order to make it more obvious of where to put documentation files
- There is now a custom landing page for the docs
- `examples` directory has been dissolved and code has been split up based on what it actual is (integration test, utility or actual example). Real example code now lives under `doc/examples`.
- All doc examples are compiled by default, ensuring the examples won't get so out of date that they fail to compile

image


doc/CMakeLists.txt
```diff
@@ -5,114 +5,108 @@ add_custom_target_subproject(sequant doc)
...
+        list(APPEND SPHINX_MISSING_MODULES ${module})
+    endif()
+endforeach()
+
+if (SPHINX_MISSING_MODULES)
+    message(FATAL_ERROR
+            "The following Python modules required for Sphinx documentation are missing:\n"
+            "  ${SPHINX_MISSING_MODULES}\n"
+            "Run the following command to install them:\n"
+            "  '${Python_EXECUTABLE} -m pip install -r ${CMAKE_CURRENT_SOURCE_DIR}/sphinx/requirements.txt'\n"
```
ajay-mk, line 84:
```suggestion
            "  '${Python_EXECUTABLE} -m pip install -r ${CMAKE_CURRENT_SOURCE_DIR}/.sphinx/requirements.txt'\n"
```

Krzmbrzl: Yeah, I notices that one just now as well :D

doc/index.rst
```diff
@@ -0,0 +1,81 @@
...
+
+SeQuant: Symbolic Tensor Algebra in C++
+=======================================
+
+SeQuant is a framework for performing symbolic algebra of tensors over scalar fields (regular tensors) and over operator fields (tensor operators in,
+e.g., quantum many-body physics). In addition to symbolic manipulation it can numerically evaluate (with an appropriate external tensor backend)
+general tensor algebra expressions.
+
+Computer algebra systems (CAS) like SeQuant are typically implemented within generic CAS like Mathematica or Maple, or using high-level languages like
+Python. In fact, version 1 of SeQuant was written in Mathematica. However, the performance of high-level languages not sufficient for practical use
```
evaleev, line 35:
```suggestion
Python. In fact, version 1 of SeQuant was written in Mathematica. However, the performance of high-level languages is not sufficient for many practical use
```

evaleev (changes requested):
few minor things, looks good otherwise, thanks!

Krzmbrzl:
There are a couple of things that still need doing (porting the remaining parts of `using.rst` from Markdown to ReST and checking whether we can implement an option to expand code snippets to show full, compilable code) but I guess as a general proof of concept of how one could do things, it suffices. And everything else needs more work anyhow, so might as well merge in this state (imo).

tests/integration/CMakeLists.txt
```diff
@@ -0,0 +1,61 @@
+set(TEST_CASES
+    # Single-Reference Coupled-Cluster equation generation (spin-orbital)
```
ajay-mk, line 2:
```suggestion
    # Single-Reference Coupled-Cluster equation generation (spin-orbital and spin-free)
```

.github/workflows/docs.yml
ajay-mk, line 1: Need to update `publish_dir` in the workflow. Should be `${{github.workspace}}/build/doc/html`

doc/developer/docs_advanced.rst
ajay-mk, line 1: Folder paths need to be updated here

doc/user/getting_started/docs.rst
```diff
@@ -1,18 +1,22 @@
...
-SeQuant uses a combination of `Doxygen <https://www.doxygen.nl/>`_, `Sphinx <https://www.sphinx-doc.org/en/master/>`_ and `Breathe <https://breathe.readthedocs.io/en/latest/>`_ to build documentation. For building the documentation locally, configure CMake with `-DSEQUANT_BUILD_DOCS=ON`.
+SeQuant uses a combination of `Doxygen <https://www.doxygen.nl/>`_, `Sphinx <https://www.sphinx-doc.org/en/master/>`_ and
+`Breathe <https://breathe.readthedocs.io/en/latest/>`_ to build documentation. For building the documentation locally, configure CMake with
+`-DSEQUANT_BUILD_DOCS=ON`.
 
 Dependencies
 ------------------------------------------------
-Building the documentation requires several Python packages. You can install them using pip. The required packages are listed in the ``requirements.txt`` file in the ``doc/sphinx`` directory. You can install them using the following command:
+Building the documentation requires several Python packages. You can install them using pip. The required packages are listed in the
+``requirements.txt`` file in the ``doc/sphinx`` directory. You can install them using the following command:
```
ajay-mk, line 11:
```suggestion
``requirements.txt`` file in the ``doc/.sphinx`` directory. You can install them using the following command:
```

doc/user/getting_started/using.rst
```diff
@@ -0,0 +1,396 @@
...
+This is achieved by the following SeQuant program:
+
+.. literalinclude:: /examples/user/getting_started/wick.cpp
+   :language: cpp
+   :start-after: start-snippet-1
+   :end-before: end-snippet-1
+   :dedent: 2
+
+.. hint:: 
+   All *core* user-facing SeQuant code lives in C++ namespace :code:`sequant`, and the shwon code assumes this namespace has been imported via
```
ajay-mk, line 26:
```suggestion
   All *core* user-facing SeQuant code lives in C++ namespace :code:`sequant`, and the shown code assumes this namespace has been imported via
```

cmake/modules/FindSphinx.cmake
ajay-mk, line 1: Since this was removed, CMake couldn't find `sphinx-build` until I installed `sphinx-cmake` via pip. It might be helpful to show a warning suggesting users install it. The Python dependency check should catch this, but the `find_package(Sphinx REQUIRED)` call comes before that.

doc/index.rst
```diff
@@ -0,0 +1,81 @@
+.. toctree::
+   :caption: User
+   :maxdepth: 2
+   :hidden:
+
+   Getting started <user/getting_started/index>
+   Installation <user/getting_started/installing>
+   Guide <user/guide/index>
```
ajay-mk, line 8: Maybe 'User guide'?

ajay-mk (commented):
I built this locally, looks great! I have a few comments.

cmake/modules/FindSphinx.cmake
Krzmbrzl: I moved the find call after the requirements check. However, in cases where the requirements have been installed except for the `sphinx-cmake` package, the check won't gain anything as there current way of checking whether a given dependency is installed doesn't work for this (thus it is filtered out).

But I guess this is fine as it won't affect new setups - only ones migrating from the prior set of requirements

doc/index.rst
Krzmbrzl: I was thinking about that, but wasn't sure if having `User Guide` under the heading `User` might be too much duplication. I added it now though, so let's see whether it bothers us going forward :)

PR #276, opened 2025-03-21
Implement `abs()` for `sequant::rational`

ajay-mk:


SeQuant/core/rational.hpp
```diff
@@ -174,6 +174,10 @@ inline std::wstring to_wolfram(const rational& t) {
            to_wolfram(denominator(t)) + L"]";
 }
 
+/// @brief Returns the absolute value of a rational number
+/// @param r a sequant::rational object
+inline rational abs(const rational& r) { return numerator(r) < 0 ? -r : r; }
```
Krzmbrzl, line 179: https://github.com/boostorg/multiprecision/issues/423#issuecomment-1039565797 suggests to me that there should already exist an `abs` implementation for Boost multiprecision types. If that is true, we should use that rather than writing our own :eyes:

ajay-mk: Ah, will close this. Just `abs()` should work for `rational` and `Complex<rational>`

PR #278, opened 2025-03-23
disable eval tracing by default even if configured with `SEQUANT_EVAL_TRACE=ON`

evaleev:
amends https://github.com/ValeevGroup/SeQuant/pull/277


Krzmbrzl:
Doesn't this make the `SEQUANT_EVAL_TRACE` option kinda meaningless or even misleading?

evaleev:
the option should go away and the tracing support is to be enabled programmatically ... typically we want to trace a block of code only

PR #285, opened 2025-03-27
Scaled leaf tensor intermediate does not have leaf tensor's label.

bimalgaudel:
eg.
before: `(42 * t{a1;i1}) -> t{a1;i1}`
after: `(42 * t{a1;i1}) -> I{a1;i1}`

nb. `I` is a dummy label.


Krzmbrzl:
Closing in favor of #286

@bimalgaudel let me know if you kept this open on purpose. If so, I'll re-open

PR #289, opened 2025-04-02
Restore docs sections

ajay-mk:
Populates `index.rst` files with sections and fixes some paths

image


doc/user/getting_started/docs.rst
```diff
@@ -1,4 +1,4 @@
-Building the Documentation
```
Krzmbrzl, line 1: To me, having the `the` in there looks/sounds better :shrug:

doc/user/getting_started/index.rst
```diff
@@ -3,3 +3,4 @@ Getting started
 
 .. toctree::
    using
+   docs
```
Krzmbrzl, line 6: I don't think this should be part of the getting started section. I'd put it into the user guide instead. After all, if someone is looking at these docs, they clearly have  pre-built documentation available so I don't think that one of the first things they want to know is how to build the docs themselves :thinking:

ajay-mk: I agree. I thin we don't need two separate pages for documentation about documentation. I have combined both and moved it to the developer section. I can move it to the user section if necessary, but since the page includes information about contributing to the documentation, I think the developer section is more appropriate.

Krzmbrzl: Sounds good. In that case, this link has to be removed though as it is now not pointing to anything, is it?

ajay-mk: Yes, I missed that. Removed.

Krzmbrzl (approved):

PR #290, opened 2025-04-02
Roll back to `TensorNetworkV1`.

bimalgaudel:
`TensorNetworkV2` needs more testing. Using it produces incorrect result for spin-orbital CC calculations in `mpqc4` when cache management is involved. This suggests the phase computation (?) in `TensorNetworkV2::canonicalize_slots` is not correct. `TensorNetworkV2::canonicalize` seems OK as it is also used heavily in closed-shell CC tests and those pass.

### DIFF
--- a/SeQuant/core/eval_expr.cpp
+++ b/SeQuant/core/eval_expr.cpp
@@ -9,7 +9,7 @@
 #include <SeQuant/core/index.hpp>
 #include <SeQuant/core/tensor.hpp>
 #include <SeQuant/core/tensor_canonicalizer.hpp>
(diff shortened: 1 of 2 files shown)


Krzmbrzl:
Please add a test case for an expression in which you observe incorrect behavior. Otherwise, it is close to impossible to fix.

Krzmbrzl:
Seems like there haven't been any test cases for phase detection. Hence, my faulty detection implementation never got caught while I was relying on test cases passing meaning that things works as expected xD

Anyway, the issue is fixed via https://github.com/ValeevGroup/SeQuant/pull/293

PR #292, opened 2025-04-06
Restore more docs

ajay-mk:
- [x] Complete porting `using.rst`
- [ ] Fix `installing.rst`


doc/examples/user/getting_started/ccd_pedestrian.cpp
Krzmbrzl, line 1: I only know the word "pedestrian" in the context of it being a person walking around on foot. Given the use here, I assume there is another meaning to the word though?

doc/examples/user/getting_started/index_space_registry_2.cpp
Krzmbrzl, line 1: Maybe it would make sense to combine both index space registry examples into a single cpp file? You could create two functions within that file for the two cases.

Imo this would facilitate manual browsing of examples.

doc/user/getting_started/installing.rst
```diff
@@ -1,2 +1,101 @@
...
+
+* `CMake <https://cmake.org/>`_ 3.27 or later
+* `C++20 compatible compiler <https://en.cppreference.com/w/cpp/compiler_support#cpp20>`_
+* `Boost <https://www.boost.org/>`_ 1.81 or later
+* `Range-V3 <https://github.com/ericniebler/range-v3.git>`_ 0.12.0 or later
+* `libperm <https://github.com/Krzmbrzl/libPerm>`_
+
+.. note:: SeQuant can download and build Boost if configured with :code:`Boost_FETCH_IF_MISSING=ON`, but the use of Boost provided by the system package manager is recommended. The following non-header-only Boost libraries are required, hence Boost must be configured/built:
+
+    * :code:`Boost.Regex`
```
Krzmbrzl, line 29: As an aside: Why do we use Boost for RegEx instead of the RegEx capabilities of the C++ standard library?

doc/user/getting_started/installing.rst
```diff
@@ -1,2 +1,101 @@
...
+----------------------
+
+.. list-table::
+   :widths: 20 10 70
+   :header-rows: 1
+
+   * - CMake Variable
+     - Default
+     - Description
+   * - `BUILD_TESTING <https://cmake.org/cmake/help/latest/module/CTest.html>`_
```
Krzmbrzl, line 51: I would recommend to instead advertise the SeQuant-specific option instead:
https://github.com/ValeevGroup/SeQuant/blob/e01d2c1cefbc37acedbbe448eb81f4d49384f14d/tests/CMakeLists.txt#L1

doc/user/getting_started/using.rst
```diff
@@ -96,286 +96,139 @@ reference state, respectively. To simplify symbolic manipulation of such express
...
 
-    Note that:
+It is also necessary to specify the *complete* space (union of all base spaces) so that the the space of unoccupied SP states can be determined:
 
-    - the tilde in $\tilde{a}$ denotes normal order with respect to single-product vacuum, and
-    - Einstein summation convention is implied, i.e., indices that appear twice in a given product (once in superscript, once in a subscript) are summed over.
+.. code-block:: c++
 
-    ## Operators
+    isr.add(L"y", 0b01, is_vacuum_occupied)
```
Krzmbrzl, line 136: Imo we should avoid having code snippets inlined into the documentation. Instead, we should include them from the examples as the latter can be compiled to verify they don't use outdated syntax etc.

doc/user/getting_started/using.rst
Krzmbrzl, line 1: I wonder whether it might be beneficial to split this single file into multiple files, so we have one file per topic. I would tend to believe that this might make navigation a bit easier :thinking:

Krzmbrzl (changes requested):

doc/examples/user/getting_started/ccd_pedestrian.cpp
ajay-mk: I think the phrase "pedestrian way" refers to something ordinary, with the hint that there is likely a better way to approach the same thing. 

I was just following what was in the Markdown document 😅

doc/user/getting_started/using.rst
ajay-mk: Maybe we can split `using.rst` to `wick.rst`, `index_spaces.rst` and `operators.rst`?

Krzmbrzl: Yeah, sounds good :+1:

doc/user/getting_started/installing.rst
ajay-mk: Addressed in [`aeba081`](https://github.com/ValeevGroup/SeQuant/pull/292/commits/aeba08168bbbe0016edd0d638341f399c725f1bb)

doc/examples/user/getting_started/index_space_registry_2.cpp
ajay-mk: Fixed in [`407643f`](https://github.com/ValeevGroup/SeQuant/pull/292/commits/407643f1e5be467125b3798d084c583df51851dc)

doc/user/getting_started/installing.rst
ajay-mk: Marking as resolved based on the Zoom conversation

doc/user/getting_started/using.rst
ajay-mk: Done, removed `using.rst` and updated `user/getting_started/index.rst`.

doc/user/getting_started/using.rst
ajay-mk: Addressed in [`aeb8b5b`](https://github.com/ValeevGroup/SeQuant/pull/292/commits/aeb8b5b0242c350829da43b2355a7f90337f38b7)

ajay-mk:
@Krzmbrzl I saw that `utilities/external-interface` has some other dependencies. Should we document those in `installing.rst`? They could also go in the specific docs for that section if you are planning to write those.

doc/user/getting_started/operators.rst
```diff
@@ -0,0 +1,92 @@
...
+   :start-after: start-snippet-3
+   :end-before: end-snippet-3
+   :dedent: 2
+
+Note the use of :code:`simplify` to rewrite an expression in a simpler form. Its role will be emphasized later.
+Unfortunately, we immediately run into the limitation of the "pedestrian" approach. Namely, the double commutator cannot be correctly obtained as
+
+.. code-block:: c++
+
+    auto c_htt = ex<Constant>(rational(1, 2)) * commutator(commutator(H, t2), t2);
```
Krzmbrzl, line 45: Maybe this should also go into the `ccd_pedestrian.cpp` file and be included as a snippet?

Krzmbrzl (approved):

Krzmbrzl:
> I saw that utilities/external-interface has some other dependencies. Should we document those in installing.rst? They could also go in the specific docs for that section if you are planning to write those.

Given that these dependencies are downloaded automatically as needed, I don't think explicit documentation is strictly necessary. Plus, the entire `external-interface` sub-project is more of a patchwork for us to be able to use SeQuant for post-processing only. Not sure if it's useful enough for other SeQuant users to warrant actual documentation... Maybe I'll write some at some point but I think for now we can just ignore it :thinking:

doc/user/getting_started/operators.rst
ajay-mk: Okay, will fix.

PR #297, opened 2025-04-23
Improve Biorthogonalization

ABesharat:
Adding new functions for closed-shell spin-tracing to get the most compact equations


Krzmbrzl:
Is the idea here to obtain the spintraced equations in the biorthogonal basis from the beginning instead of doing the transformation after spintracing?

ABesharat:
That could be an idea to consider. Here, we still do biorthogonal transformation after spintracing. The difference is that for spintracing, we do full expansion instead of partial ( in terms of S) expansion. So, after spintracing, we have all of the terms. Then we do biorthogonal transformation, then apply hash filtering to get the most compact set of equations. However, this approach is not optimal, because biorthogonal transformation still tries to add permutation-related terms that we already have. So, it does unnecessary work, while we just need the scaling operation.

evaleev:
when making changes to the core functionality it's a good idea to check the performance of basic benchmarks that we have thanks to @Krzmbrzl ... it seems that canonicalizer tests are slower than they should be, at least on my laptop (the data should be directly comparable to V1 data in #303 ):
```
canonicalize/1                                                     107 ns          107 ns      6496218
canonicalize/2                                                   20265 ns        20263 ns        34575
canonicalize/3                                                   32635 ns        32626 ns        21464
canonicalize/4                                                   40275 ns        40271 ns        17204
canonicalize/5                                                   44689 ns        44684 ns        15650
canonicalize/6                                                   54462 ns        54449 ns        12906
```

not sure why there is perf regression

ABesharat:
It is working now, but to be merged I want to apply some changes. Such as discarding all the debugging, changing vector to small vector. Also, I need to move cleanup from mpqc4 to SeQuant. Now it is working for triples and quadruples, I should generalize it.

Sent from my T-Mobile 5G Device

SeQuant/domain/eval/eval.hpp
```diff
@@ -535,6 +537,36 @@ ResultPtr evaluate_antisymm(Args&&... args) {
   }
   return result;
 }
+
+///
+/// \brief Calls @code evaluate followed by the biorthogonal-cleanup function.
```
evaleev, line 542: a couple of issues:
- each `@code` need a `@endcode` to close code block. To highlight a single word use back quotes
- this needs a more details dox. Ideally you can show the expression for the triples or refer to an equation in the paper (make sure to give DOI if referring to the paper)

to make sure documentation looks good you need to generate documentation and check it out; CI does not ensure that documentation looks correct, only that it builds

evaleev: if the biorthogonal-cleanup is documented elsewhere then can just refer to that instead of giving details here. Ideally there is one place where the cleanup is described.

SeQuant/domain/mbpt/spin.cpp
```diff
@@ -932,23 +935,82 @@ ExprPtr S_maps(const ExprPtr& expr) {
...
+  // hash filtering logic for R > 2
+  container::map<std::size_t, container::vector<ExprPtr>> largest_coeff_terms;
+
+  for (const auto& term : *expr) {
+    if (!term->is<Product>()) continue;
+
+    auto product = term->as<Product>();
+    auto scalar = product.scalar();
+
+    sequant::TensorNetworkV2 tn(product);
```
evaleev, line 953: we now use TNV3 everywhere

evaleev (changes requested):

evaleev (approved):

PR #300, opened 2025-05-04
fix: Handle empty range in `input_to_canonical_parity` function

ajay-mk:
EOM-IP/EA calculations failing because this function encountered empty index range.


Krzmbrzl:
Is there a specific reason why you are using TNv1 instead of TNv2?

evaleev (approved):

ajay-mk:
> Is there a specific reason why you are using TNv1 instead of TNv2?

My MPQC branch was pinned to an older SeQuant commit which was using TNv1 in eval code. I have updated it, and now it uses TNv2.

PR #305, opened 2025-05-20
optimized canonicalization

ABesharat:
trying to optimize canonicalization of sum

### DIFF
--- a/SeQuant/core/expr.cpp
+++ b/SeQuant/core/expr.cpp
@@ -6,10 +6,10 @@
 #include <SeQuant/core/algorithm.hpp>
 #include <SeQuant/core/expr.hpp>
 #include <SeQuant/core/logger.hpp>
+#include <SeQuant/core/runtime.hpp>
 #include <SeQuant/core/tensor.hpp>
 #include <SeQuant/core/tensor_canonicalizer.hpp>
 #include <SeQuant/core/tensor_network_v2.hpp>
-
 #include <range/v3/all.hpp>
 
 #include <thread>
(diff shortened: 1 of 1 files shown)


SeQuant/core/expr.cpp
```diff
@@ -6,12 +6,13 @@
 #include <SeQuant/core/algorithm.hpp>
 #include <SeQuant/core/expr.hpp>
 #include <SeQuant/core/logger.hpp>
-#include <SeQuant/core/runtime.hpp>
```
evaleev: revert

*Rationale*
insignificant changes like reordering of `#include` statements without good reason should be avoided. `#include`s should be organized into space-separated groups (so that clang-format can sort alphabetically within each group) and ordered as follows:

- forward declaration headers (`fwd.h`, etc.)
- headers of the current package
- headers of dependent packages, from more specialized to less specialized, one group per package
- standard headers

SeQuant/core/expr.cpp
```diff
@@ -341,8 +342,7 @@ void Sum::adjoint() {
                       });
   *this = Sum(ranges::begin(adj_summands), ranges::end(adj_summands));
 }
-#define canonicalization
-#ifdef canonicalization
```
evaleev: why is this macro needed? macros should be avoided whenever possible

SeQuant/core/expr.cpp
```diff
@@ -350,85 +352,88 @@ ExprPtr Sum::canonicalize_impl(bool multipass) {
...
     }
+    summands_.swap(new_summands);
 
     if (Logger::instance().canonicalize)
       std::wcout << "Sum::canonicalize_impl (pass=" << pass
                  << "): after reducing summands = "
                  << to_latex_align(shared_from_this()) << std::endl;
   }
 
-  return {};  // side effects are absorbed into summands
```
evaleev, line 431: revert, the comment was useful

evaleev (changes requested):

Krzmbrzl:
I assume this was superseded by #314 ?

evaleev (approved):

PR #312, opened 2025-06-26
Include cstdlib header and find catch2 before fetching.

bimalgaudel:


cmake/modules/FindOrFetchCatch2.cmake
```diff
@@ -1,3 +1,5 @@
+find_package(Catch2)
```
ajay-mk, line 1: VRGFindOrFetchPackage module should do this. It tries to use find_package first, and uses FetchContent if not found. 
See https://github.com/ValeevGroup/kit-cmake/blob/master/modules/VRGFindOrFetchPackage.cmake

evaleev (approved):

cmake/modules/FindOrFetchCatch2.cmake
```diff
@@ -13,6 +13,8 @@ if (NOT TARGET Catch2::Catch2)
 
 	if (Catch2_SOURCE_DIR)
 		list(APPEND CMAKE_MODULE_PATH "${Catch2_SOURCE_DIR}/extras")
+  	elseif(Catch2_DIR)
```
evaleev, line 16: the commit mentions using vcpkg-provided Catch2 but this block of code is in the from-source clause. Probably needs to be moved outside?

Issue #172, opened 2024-01-02
Make Index::LabelCompare do numeric comparison on the subscript (numeric) part of an index.


bimalgaudel:
As of now:
`Index{L"i_10"} < Index{L"i_2"}` is true.

After:
`Index{L"i_10"} > Index{L"i_2"}` is true.

While equality (==) (or !=) holds with either string comparison or numeric comparison, ordering/sorting is likely to introduce a bug with string comparison alone, as the programmer naturally might expect a numeric comparison on the numeric portion.

Krzmbrzl:
The symptoms have been fixed in #177 but imo indices should be refactored to contain a label (string) and an ID (integer) as two separate members. Then comparing the ID numerically is easy.
Where needed the two components could be merged into a single string very easily (a easier than it is to disassemble the string into the two components)

PR #315, opened 2025-07-07
faster Index

evaleev:
no longer uses label for representation, instead space + ordinal are stored explicitly to speed up comparisons/copies.

WARNING: due to the change in representation hash value changed, which breaks several hash-dependent tests.

### DIFF
--- a/SeQuant/core/utility/indices.hpp
+++ b/SeQuant/core/utility/indices.hpp
@@ -267,16 +267,15 @@ TensorOfTensorIndices<Container> tot_indices(Rng const& idxs) {
 }
 
 ///
-/// Does the numeric comparison of the index suffixes using less-than operator.
+/// Does the numeric comparison of the index ordinals using less-than operator.
 ///
 /// \param idx1
 /// \param idx2
-/// \return True if the numeric suffix of \c idx1 is less than that of \c idx2.
+/// \return True if the ordinal of \c idx1 is less than that of \c idx2. Since
+/// ordinals are optional, the semantics is that of std::optional
 ///
 inline bool ordinal_compare(Index const& idx1, Index const& idx2) {
-  auto ord1 = idx1.ordinal();
-  auto ord2 = idx2.ordinal();
-  return (ord1 && ord2) && ord1.value() < ord2.value();
+  return idx1.ordinal() < idx2.ordinal();
 }
 
 ///


Krzmbrzl:
FWIW, I'm pretty sure this fixes https://github.com/ValeevGroup/SeQuant/issues/172 (at least it should)

SeQuant/core/index.hpp
```diff
@@ -79,62 +88,137 @@ class Index : public Taggable {
...
-  /// convertible into an IndexSpace (@sa IndexSpace::instance )
-  template <typename String,
+  /// convertible into an IndexSpace (@sa IndexSpace::instance ); if @p label
+  /// is an rvalue reference will scavenge it for the label of the new index
+  /// @warning nonexplicit to make Index creation as ergonomic as possible, at the cost of
+  /// slightly increased danger
+  template <typename String, typename Integer = std::uint64_t,
             typename = std::enable_if_t<meta::is_basic_string_convertible_v<
-                std::remove_reference_t<String>>>>
+                std::remove_reference_t<String>> && std::is_integral_v<Integer>>>
```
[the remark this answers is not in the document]
evaleev: resolved via 5380af26a7902e177db4902f3e5baa4fce9b25cf

Krzmbrzl:
Final remark: Do we really need index label memoization? I'd argue that during processing nothing should make use of the label anyway (to avoid string manipulation instead of working with space + ordinal directly) and for stringification, I doubt that memoization yields useful performance improvements.

So I'm wondering whether making the index object larger (byte-wise) is actually worth it :thinking:

evaleev:
> Final remark: Do we really need index label memoization? I'd argue that during processing nothing should make use of the label anyway (to avoid string manipulation instead of working with space + ordinal directly) and for stringification, I doubt that memoization yields useful performance improvements.
> 
> So I'm wondering whether making the index object larger (byte-wise) is actually worth it 🤔

do not have the measurements. p.s. and memoization makes label/full_label non-reentrant. so perhaps might make sense. but: small steps.

SeQuant/core/index.hpp
```diff
@@ -352,11 +424,23 @@ class Index : public Taggable {
...
+  /// @return the memoized label as a UTF-8 encoded wide-character string
   /// @note label format is `base` or `base_ordinal`
   /// @warning this does not include the proto index labels, use
   /// Index::full_label() instead
-  std::wstring_view label() const { return label_; }
+  std::wstring_view label() const {
+    if (!label_) {
+      label_ = space_.base_key();
+      if (ordinal_ != default_ordinal)
+        *label_ += L'_' + std::to_wstring(ordinal_);
```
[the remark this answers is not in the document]
Krzmbrzl: I don't see it as weird. To me this looks like a reasonable default :shrug: 

> Forbidding ordinal-free labels is plain confusing.

I don't really see why that should be the case. SeQuant simply provides a syntax for specifying indices and that includes an ordinal. That could simply be how things are. I wouldn't consider it more or less confusing as if someone wants to use `i1` in the `Index` ctor (which to my knowledge is not supported) or any other kind of syntax.

If anything, I think it is confusing that the same string can be either an `IndexSpace` or an `Index` which clearly are quite different things.

Overall, I don't really have a strong opinion on this so feel free to mark as resolved in order to move on.

SeQuant/core/index.hpp
```diff
@@ -79,62 +110,173 @@ class Index : public Taggable {
...
 
-  /// @param label the label, does not need to be unique
+  /// copy constructor
+  /// @warning memoized data (label, full_label) is not copied
+  Index(const Index &idx) noexcept
+      : Taggable(idx),
+        space_(idx.space_),
+        ordinal_(idx.ordinal_),
+        proto_indices_(idx.proto_indices_),
+        symmetric_proto_indices_(idx.symmetric_proto_indices_) {}
```
Krzmbrzl, line 120: I think I would implement the copy ctor in terms of the copy-assignment ctor in order to reduce code duplication.
```suggestion
  Index(const Index &idx) noexcept { *this = idx; }
```

SeQuant/core/index.hpp
```diff
@@ -79,62 +110,173 @@ class Index : public Taggable {
...
+  }
+
+  /// @param space (a const ref to) the IndexSpace object that specifies to this
+  /// space this object belongs
+  /// @param ord the index ordinal
+  template <meta::integral I>
+  Index(const IndexSpace &space, std::optional<I> ord)
+      : space_(space), ordinal_(ord) {
+    check_nonreserved();
+  }
```
Krzmbrzl, line 162: Can't we use constructor-chaining (delegating constructors) to avoid duplication of constructor body definitions?
I guess this goes somewhat in the direction of Copilot's review comment

I think this applies to most constructors here

SeQuant/core/index.hpp
```diff
@@ -79,62 +110,173 @@ class Index : public Taggable {
...
-  Index(String &&label, const IndexSpace &space, IndexList proto_indices,
+  Index(const IndexSpace &space, container::vector<Index> proto_indices,
         bool symmetric_proto_indices = true)
-      : label_(to_wstring(std::forward<String>(label))),
-        space_(space),
-        proto_indices_(proto_indices),
-        symmetric_proto_indices_(symmetric_proto_indices) {
+      : Index(space) {
+    proto_indices_ = std::move(proto_indices);
+    symmetric_proto_indices_ = symmetric_proto_indices;
```
Krzmbrzl, line 174:
```suggestion
      : Index(space), proto_indices_ = std::move(proto_indices), symmetric_proto_indices_ = symmetric_proto_indices {
```

SeQuant/core/index.hpp
```diff
@@ -79,62 +110,173 @@ class Index : public Taggable {
...
   /// @param label the index label, does not need to be unique, but must be
-  /// convertible into an IndexSpace (@sa IndexSpace::instance )
+  /// convertible into an IndexSpace (@sa IndexSpace::instance ); if @p label
+  /// is an rvalue reference will scavenge it for the label of the new index
+  /// @warning nonexplicit to make Index creation as ergonomic as possible, at
+  /// the cost of slightly increased danger
   template <typename String,
-            typename = std::enable_if_t<meta::is_basic_string_convertible_v<
-                std::remove_reference_t<String>>>>
+            typename = meta::EnableIfAllBasicStringConvertible<String>>
```
Krzmbrzl, line 268: We should turn this into a concept and then use a`requires` clause. That will yield much nicer error messages

SeQuant/core/space.hpp
```diff
@@ -464,19 +465,26 @@ class IndexSpace {
 inline const IndexSpace IndexSpace::null;
 inline const IndexSpace::Attr IndexSpace::Attr::null;
 
-/// @return true if type2 is included in type1, i.e. intersection(type1, type2)
-/// == type2
+inline auto hash_value(const IndexSpace &space) {
+  auto result = hash_value(static_cast<std::int64_t>(space.attr()));
+  hash::combine(result, space.base_key());
+  hash::combine(result, space.approximate_size());
```
Krzmbrzl, line 471: I don't think we'll want the size to be part of the hash. Imo, having two index spaces that differ only in their size doesn't make any sense :thinking: 

And I think the same argument kinda holds for the base label as well

This is particularly true as these properties are not taken into account in `IndexSpace::operator==` and I believe we should definitely match the hashing with identity checking

SeQuant/core/utility/indices.hpp
```diff
@@ -273,10 +273,10 @@ TensorOfTensorIndices<Container> tot_indices(Rng const& idxs) {
...
 /// \return True if the numeric suffix of \c idx1 is less than that of \c idx2.
 ///
-inline bool suffix_compare(Index const& idx1, Index const& idx2) {
-  auto&& s1 = idx1.suffix();
-  auto&& s2 = idx2.suffix();
-  return (s1 && s2) && s1.value() < s2.value();
+inline bool ordinal_compare(Index const& idx1, Index const& idx2) {
+  auto ord1 = idx1.ordinal();
+  auto ord2 = idx2.ordinal();
+  return (ord1 && ord2) && ord1.value() < ord2.value();
```
Krzmbrzl, line 279: Any specific reason as to not simply relying on optional's `operator<` (that also covers the cases of null values)?

This implementation treats any index with suffix as equal to any index without a suffix. Is this really what we want? :thinking:

tests/unit/test_mbpt.cpp
```diff
@@ -401,16 +401,13 @@ TEST_CASE("mbpt", "[mbpt]") {
...
-          L"{{{\\frac{1}{4}}}{\\bar{R}^{{i_1}{i_2}}_{{a_1}{a_2}}}{\\tilde{a}^"
-          L"{{a_1}{"
-          L"a_2}}_{{i_1}{i_2}}}}\\bigr) }");
+      REQUIRE(to_latex(R33) ==
+              L"{ "
+              L"\\bigl({{{\\frac{1}{36}}}{\\bar{R}^{{i_1}{i_2}{i_3}}_{{a_1}{a_"
+              L"2}{a_3}}}{\\tilde{a}^{{a_1}{a_2}{a_3}}_{{i_1}{i_2}{i_3}}}} + "
+              L"{{{\\frac{1}{4}}}{\\bar{R}^{{i_1}{i_2}}_{{a_1}{a_2}}}{\\tilde{"
+              L"a}^{{a_1}{a_2}}_{{i_1}{i_2}}}} + "
+              L"{{R^{{i_1}}_{{a_1}}}{\\tilde{a}^{{a_1}}_{{i_1}}}}\\bigr) }");
```
Krzmbrzl, line 410: I would recommend to use the matcher expressions I introduced in other test cases in order to have tests that are independent of the canonical form (and also easier to read for humans).

tests/unit/test_spin.cpp
```diff
@@ -676,10 +676,11 @@ SECTION("Closed-shell spintrace CCD") {
...
           closed_shell_CC_spintrace(pno_ccd_energy_so_as_sum);
       REQUIRE_THAT(pno_ccd_energy_sf,
-                   SimplifiesTo("2 g{a1<i1,i2>,a2<i1,i2>;i1,i2}:N-C "
-                                "t{i1,i2;a1<i1,i2>,a2<i1,i2>}:N-C - "
+                   SimplifiesTo("- "
                                 "g{a1<i1,i2>,a2<i1,i2>;i1,i2}:N-C "
-                                "t{i1,i2;a2<i1,i2>,a1<i1,i2>}:N-C"));
+                                "t{i1,i2;a2<i1,i2>,a1<i1,i2>}:N-C"
+                                " + 2 g{a1<i1,i2>,a2<i1,i2>;i1,i2}:N-C "
+                                "t{i1,i2;a1<i1,i2>,a2<i1,i2>}:N-C "));
```
Krzmbrzl, line 683: Was this test case failing? I would have expected that the matcher expression should have taken care of any changes in the canonical form (here: order of summands)?

Krzmbrzl (commented):
I think a test case for `i_11 > i_5` should be added as well

tests/unit/test_spin.cpp
evaleev: see comments re: hashing. This did expose some leakage of side effects from unit tests and is also fixed in this PR.

tests/unit/test_mbpt.cpp
evaleev: I agree ...

SeQuant/core/index.hpp
evaleev: delegating ctor can't be supplemented by field initialization

SeQuant/core/index.hpp
evaleev: in my experience this will require quite a bit of work and will be harder to read, although the flip side is more verbose code.

SeQuant/core/index.hpp
evaleev: I don't know if the compiler will make this as efficient as the explicit version. For core concepts like Index I don't mind being verbose

Krzmbrzl: I would assume that the code will be equally efficient. But alright :+1:

evaleev: I would certainly hope so, but strictly speaking default initialization will occur first, then the data will be overwritten by copy assignment.

SeQuant/core/space.hpp
```diff
@@ -489,13 +497,12 @@ inline bool includes(const IndexSpace &space1, const IndexSpace &space2) {
...
-/// IndexSpace are equal if they have equal @c IndexSpace::type(),
-/// @c IndexSpace::qns(), and @c IndexSpace::base_key().
+/// IndexSpace are equal if they have equal @c IndexSpace::type() and
+/// @c IndexSpace::qns().
 ///
 [[nodiscard]] inline constexpr bool operator==(
     IndexSpace const &space1, IndexSpace const &space2) noexcept {
-  return space1.type() == space2.type() && space1.qns() == space2.qns() &&
-         space1.base_key() == space2.base_key();
+  return space1.type() == space2.type() && space1.qns() == space2.qns();
```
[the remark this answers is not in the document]
evaleev: fixed

SeQuant/core/index.hpp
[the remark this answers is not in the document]
evaleev: past experience says someone will say "SeQuant is stupid, it won't let me choose index labels without defining a vocabulary first". Most will be happy with the vocabulary.

PR #317, opened 2025-07-07
Compute intermediate hashes strictly.

bimalgaudel:
This is required instead of lazily computing them during binarization to avoid surprises. See the test cases.

### DIFF
--- a/SeQuant/core/eval_expr.cpp
+++ b/SeQuant/core/eval_expr.cpp
@@ -324,10 +324,10 @@ EvalExprNode binarize(Product const& prod) {
                  | ranges::to_vector;
 
   auto hvals = factors | transform([](auto&& n) { return n->hash_value(); });
+  auto const hs = imed_hashes(hvals) | ranges::to_vector;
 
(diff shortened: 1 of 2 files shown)


evaleev:
the issue seems to be due to `imed_hashes` returning a range rather than a container ... do you need to apply this fix in the other place where `imed_hashes` used?

bimalgaudel:
Turns out the problem is in the statement:

`auto h = ranges::at(hs, ++i);`

`ranges::at` requires `hs` to be a random-access range.

In the following piece of code
```cpp
auto make_prod = [i = 0, hs = imed_hashes(hvals)](..) mutable {
     auto h = ranges::at(hs, ++i);
};
```

`hs` is not a random-access range. The compiler should have helped here but does not. However, when `i` is taken outside the lambda introduction so that `mutable` keyword is not required, the compiler can pick up the problem.

```cpp
auto i = 0;
auto make_prod = [&i, hs = imed_hashes(hvals)](..) {
     auto h = ranges::at(hs, ++i); // does not compile
};
```

Issue #316, opened 2025-07-07
Run `doc/examples/` in CI


ajay-mk:
Currently, we build the example code in `doc/examples/` as part of the CI to ensure they compile. We should extend this to also run the examples, where applicable, to catch any runtime errors.

PR #318, opened 2025-07-07
Run `/doc/examples/`

ajay-mk:
Address issue #316 

Examples files in `doc/examples` can now be tested as part of the test suite. 

We already separate examples into executables and libraries based on the presence of `main()` function. Now, the executables are added to the test suite, and `ctest` will discover them as part of `check_sequant` target. 

Time consuming tests can be disabled by using adding `// SEQUANT_DOC_EXAMPLE: DO NOT RUN` comment in the source file (see `doc/examples/cc.cpp`)

### DIFF
--- a/doc/examples/CMakeLists.txt
+++ b/doc/examples/CMakeLists.txt
@@ -8,6 +8,9 @@ endif()
 
 file(GLOB_RECURSE EXAMPLE_CODES LIST_DIRECTORIES false CONFIGURE_DEPENDS "*.cpp")
 
+# Create a list to track runnable examples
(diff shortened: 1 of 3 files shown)


ajay-mk:
image
Now captures the outdated example in CI.

ajay-mk:
The `index_spaces.cpp` file had a very old API call `vacuum_occupied_space()` to set the occupied space of the ISR. It compiled because there is another `vacuum_occupied_space()` which returns occupied space. 

Fixed in https://github.com/ValeevGroup/SeQuant/pull/318/commits/4729728b552abbe202d21ff4658800b51285daa6, also includes some changes to make sure the example runs.

PR #321, opened 2025-07-09
TiledArray code generation

bimalgaudel:
Consider the following sum of tensor networks:

$$(A(a_1 x_1) B(x_1 x_2)) C(x_2 b_1) + (A(a_1 y_1) B(y_1 y_2)) D(y_2 b_1) + (C(a_1 x_1) D(x_1 x_2)) E(x_2 b_1)$$

This PR attempts to support `TiledArray` code generation from it that looks like the following or its equivalent.

```cpp

// declare the total evaluation result
TArray R;

// first and second term share a common intermediate (A*B)
// evaluate it as part of the first term evaluation

TArray I1;
I1("a1,x2") = A("a1,x1") * B("x1,x2");
R("a1,a2")  = I1("a1,x2") * C("x2,b1");
// done evaluating first term

// reuse the intermediate from first term evaluation
R("a1,a2") += I1("a1,y2") * D("y2,b1");
// done evaluating second term

// third term shares no intermediates
// no need to split up the evaluation like for the first term
R("a1,b1") += (C("a1,x1") * D("x1,x2")) * E("x2,b1");
// done evaluating third term

```

### DIFF
--- a/tests/unit/CMakeLists.txt
+++ b/tests/unit/CMakeLists.txt
@@ -4,9 +4,10 @@ include(Catch)
 add_executable(unit_tests-sequant ${BUILD_BY_DEFAULT}
     "test_asy_cost.cpp"
     "test_binary_node.cpp"
-	"test_biorthogonalization.cpp"
+    "test_biorthogonalization.cpp"
     "test_bliss.cpp"
     "test_canonicalize.cpp"
+    "test_codegen_ta.cpp"
     "test_eval_expr.cpp"
     "test_eval_node.cpp"
     "test_export.cpp"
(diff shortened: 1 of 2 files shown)


Krzmbrzl:
I see. Does that mean there is an actual disadvantage in creating the intermediates explicitly even in cases where it's not necessary? I mean, other than code readability

But I guess the bottom line is that it can likely be done right now by adding some logic into the backend generator but it kinda goes against the design philosophy of the export framework (backend generators should contain only a minimal amount of logic in order to be easily written).

I would propose focusing on getting the export framework with its current abilities and restrictions merged first and then think of how to generalize it to support non-binary contractions as well. I think this would also be nice for generating code that is interpreted by a specialized tensor compiler.

evaleev:
I agree, minimal binarized TA code emission should be the initial goal. Dispatch to more general backends capable of evaluating whole TNs is longer term.

evaleev:
obsolete in light of ongoing export revamp

PR #323, opened 2025-07-10
Fix parenthesization in condition

Krzmbrzl:
The previous version allowed expressions that aren't tensors to be added to the list of tensors in cases where the current node represents a sum.


evaleev (approved):

Krzmbrzl:
FYI: Not only was the logic wrong, but this actually caused undefined behavior because in cases where the first condition evaluates to `false`, the second branch of the `||` is executed which will dereference the optional and contrary to `optional::value`, `optional::operator*` does **not** check for whether the value is present or not (and hence also doesn't throw an exception in cases where it doesn't contain a value).

@evaleev not sure if you knew about this, but at least to me this was new (I thought the operators had the same semantics as the `value` function). Maybe we should simply stay away from using `optional::operator*` and `optional::operator->` and instead prefer explicit `value` calls?

PR #324, opened 2025-07-10
Update tracked libPerm version

Krzmbrzl:
The old version still has some issues in the way it configures LTO


ajay-mk:
@Krzmbrzl Please update the tag in INSTALL.md too. I wonder why git hooks didn't do that.

PR #314, opened 2025-07-02
TNv2 performance improvements

Krzmbrzl:
Benchmarking indicated that the use of maps was actually a bottleneck in this function (particularly the map used for the index -> vertex mapping). In order to mitigate this, the maps have all been replaced with vector-based alternatives.

The overall effect on performance is only rather moderate (maybe up to 5%). Here's the benchmark data:
```
### New

Running ./benchmarks/sequant_benchmarks
Run on (16 X 1,399.85 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x8)
  L1 Instruction 32 KiB (x8)
  L2 Unified 512 KiB (x8)
  L3 Unified 4,096 KiB (x2)
Load Average: 1.77, 1.98, 2.45
-----------------------------------------------------------------------------
Benchmark                                   Time             CPU   Iterations
-----------------------------------------------------------------------------
random_tensor_network/0/128_cv           2.52 %          2.52 %            10
random_tensor_network/0/128_mean         1.37 ms         1.37 ms           10
random_tensor_network/0/128_median       1.38 ms         1.38 ms           10
random_tensor_network/0/128_stddev      0.035 ms        0.035 ms           10
random_tensor_network/0/16_cv            2.01 %          2.02 %            10
random_tensor_network/0/16_mean         0.068 ms        0.068 ms           10
random_tensor_network/0/16_median       0.068 ms        0.068 ms           10
random_tensor_network/0/16_stddev       0.001 ms        0.001 ms           10
random_tensor_network/0/1_cv             0.94 %          0.94 %            10
random_tensor_network/0/1_mean          0.009 ms        0.009 ms           10
random_tensor_network/0/1_median        0.009 ms        0.009 ms           10
random_tensor_network/0/1_stddev        0.000 ms        0.000 ms           10
random_tensor_network/0/256_cv           2.10 %          2.10 %            10
random_tensor_network/0/256_mean         4.67 ms         4.67 ms           10
random_tensor_network/0/256_median       4.66 ms         4.66 ms           10
random_tensor_network/0/256_stddev      0.098 ms        0.098 ms           10
random_tensor_network/0/2_cv             0.78 %          0.78 %            10
random_tensor_network/0/2_mean          0.012 ms        0.012 ms           10
random_tensor_network/0/2_median        0.012 ms        0.012 ms           10
random_tensor_network/0/2_stddev        0.000 ms        0.000 ms           10
random_tensor_network/0/32_cv            2.93 %          2.93 %            10
[...]

PR #329, opened 2025-07-11
redundant call for index replacement

ABesharat:
There is a redundant call for index replacement in canonicalize_graph, while the job can be handled later in canonicalize function

### DIFF
--- a/SeQuant/core/tensor_network_v2.cpp
+++ b/SeQuant/core/tensor_network_v2.cpp
@@ -440,15 +440,16 @@ void TensorNetworkV2::canonicalize_graph(const NamedIndexSet &named_indices) {
 
   sort_via_indices<false>(edges_, index_sorter);
 
-  for (const Edge &current : edges_) {
-    const Index &idx = current.idx();
-
-    if (!is_anonymous_index(idx)) {
-      continue;
-    }
-
-    idxrepl.insert(std::make_pair(idx, idxfac.make(idx)));
-  }
+  // there is no need to fill idxrep in canonicalize_graph. The job will be
+  // handled in canonicalize fn for (const Edge &current : edges_) {
+  //   const Index &idx = current.idx();
+  //
+  //   if (!is_anonymous_index(idx)) {
+  //     continue;
+  //   }
+  //
+  //   idxrepl.insert(std::make_pair(idx, idxfac.make(idx)));
+  // }
 
   if (Logger::instance().canonicalize) {
     for (const auto &idxpair : idxrepl) {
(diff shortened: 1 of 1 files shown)


ABesharat:
profiling for canonicalize.cpp, case 6:
V2::canonicalize_graph
	V2::canonicalize_graph::index_replacement: 1,500ns // redundant
	V2::canonicalize_graph::applyIdxReplacements: 1,250ns // redundant
V2::canonicalize	
        V2::canonicalize::index_replacement: 833ns // the job can be handled here
        V2::canonicalize::apply_index_replacements: 208ns // can be optimized when edxrepl is not empty

Krzmbrzl:
Imo we should keep the optimizations together in #314. Otherwise, it gets hard to track how different changes influence each other 👀
To that end, you can simply create a PR against my fork for the branch #314 is based on.

As for the actual changes:
> There is a redundant call for index replacement in canonicalize_graph, while the job can be handled later in canonicalize function

I don't see the redundancy. The entire job of `canonicalize_graph` is to bring the expression into a canonical form. This (to my understanding) must involve index replacements as (iirc) the subsequent tensor sorting can depend on the specific index names 🤔

Could you walk me through your reasoning in a bit more detail?

ABesharat:
In canonicalize_graph we fill idxrepl and apply index replacement (before particle swap). Then we do the same in canonicalize function.
This is the resutl for canonicalize.cpp/case 3:

canonicalize_graph: idxrepl is not empty
i_1 -> i_2
i_2 -> i_1
i_3 -> i_4
i_4 -> i_3
canonicalize: idxrepl is not empty
i_1 -> i_2
i_2 -> i_1
i_3 -> i_4
i_4 -> i_3

This seems redundant to me. If we simply comment out this step in canonicalize_graph, the job will be handled later in the second step. This is my understanding, but I might be wrong.

Sometimes idxrep is empty and we still call apply_index_replacement (like in case 5), that is why I added  if(!idxrepl.empty()). This might not be a significant optimization. Even calling it does not take a considerable time but it is avoidable:
V2::canonicalize::apply_index_replacements: 208ns

ABesharat:
Do we have such a test for spin orbitals? Because this change passed all the test we have.

Krzmbrzl:
I don't know. My main point here is that we should not skip any steps unless we can (mathematically) proof that it is safe to do. Canonicalization is really tricky to get right and if we start optimizing steps away we might end up with edge cases that don't canonicalize properly 👀

evaleev:
@ABesharat fast canonicalization is in general needed to produce "pleasant" canonical order which differs from the order produced by Bliss by trivial (local) reorderings of slots and index renamings. This is what TNv1 called fast canonicalization.  Effectively TNv2 redefined the fast canonicalization to mean the same as old_slow(TNv1) + fast (TNv1) combination.

You may not have been able to see effects of dropping some index renamings because our unit tests do not have 100% coverage in the feature space. Also, Sum::canonicalize does by default always forced slow canonicalization to be followed by fast, which for TNv1 was needed, but for TNv2 is redundant. This probably would in effect mean that with TNv2 you could drop some renamings and never the effect. Still, it's a bad idea to not perform full old_slow and fast canonicalizations.

PR #334, opened 2025-07-20
Optimize docs deployment

ajay-mk:
#### Adds `force_orphan: true` option to gh-pages deployment action.

This should reduce the repo size and improve cloning time. 

For reference: https://github.com/peaceiris/actions-gh-pages#%EF%B8%8F-force-orphan-force_orphan


ajay-mk:
> Does this mean that the `gh-pages` branches will only contain a single commit instead of a commit history containing all previous states of the doc pages?

Yes

PR #338, opened 2025-07-24
`memoizing_hash` improvements

evaleev:
- change {`Product`,`Sum`}::`memoizing_hash` so that for single {factor,summand} it produces hashes identical to the hash of the {factor,summand} itself
- `memoizing_hash` memoizes, and validates against memoized value in Debug builds

### DIFF
--- a/SeQuant/core/expr.hpp
+++ b/SeQuant/core/expr.hpp
@@ -832,7 +832,11 @@ class Constant : public Expr {
 
  private:
   hash_type memoizing_hash() const override {
-    hash_value_ = hash::value(value_);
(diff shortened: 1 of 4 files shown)


evaleev:
@Krzmbrzl what could valgrind do to change behavior of tests?

https://github.com/ValeevGroup/SeQuant/actions/runs/16506220516/job/46677408214?pr=338#step:12:120

evaleev:
It seems this is not valgrind issue but test interaction issue ... valgrind builds are configured differently, and that seems to cause this. Of course the test succeeds if run standalone

PR #344, opened 2025-07-31
Tensor::hash_value accounts for braket and particle symmetry

evaleev:


Krzmbrzl:
Iirc `Tensor::operator==` doesn't account for those symmetries. If the hash does it, so should equality checks.

PR #357, opened 2025-08-15
TNV3: support for `BraKetSymmetry`

evaleev:
this is implemented just like support for canonicalization of slots (bra and ket introduced in v3) and braket slot bundles (in v2). Need to introduce vertices for bra and ket slot bundles and swap them if needed. Currently we don't have the support for conjugation as canonicalization biproduct, but I think tensors already have the support for conjugation. For real tensors straight swap is enough.

### DIFF
--- a/SeQuant/core/abstract_tensor.hpp
+++ b/SeQuant/core/abstract_tensor.hpp
@@ -249,6 +249,10 @@ class AbstractTensor {
   virtual void _permute_braket(std::span<std::size_t> perm) {
     permute_braket_impl(_bra_mutable(), _ket_mutable(), perm);
   }
+  /// swaps bra and ket slots
+  virtual void _swap_bra_ket() {
+    throw missing_instantiation_for("_swap_bra_ket");
+  }
 
  private:
   /// @return mutable view of bra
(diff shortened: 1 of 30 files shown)


SeQuant/core/abstract_tensor.hpp
```diff
@@ -349,7 +352,7 @@ class AbstractTensor {
 /// objects.
 /// @{
 inline auto braket(const AbstractTensor& t) { return t._braket(); }
-inline auto indices(const AbstractTensor& t) { return t._braketaux(); }
+inline auto braketaux(const AbstractTensor& t) { return t._braketaux(); }
```
Krzmbrzl, line 355: Why the rename? I much prefer having a function which will give me _all_ indices instead of a function that gives me indices in the given groups (which can't be amended in case new groups get added and hence code relying on getting all indices needs to be refactored - but one has to explicitly think of that

evaleev: renamed for consistency and explicitness, the meaning is the same. We had `braket()` already. We can keep both, if you have a lot of code that uses `indices()` already.

the bigger issue is that the distinction between slots and indices is not explicit in these functions ...

evaleev: > I disagree in a general context. The meaning is the same _now_ because we don't have more index groups but as soon as we add another group, the meaning will no longer be equivalent. Exactly as happened to `braket`.

I sure hope we don't :)

evaleev: > I'd definitely appreciate keeping the `indices` function, albeit for semantic reasons rather than having a lot of code that uses it.

done

PR #363, opened 2025-08-19
Code maintenance

Krzmbrzl:
Mostly related to handling of warnings

Most relevant changes:
-  Sets `-Werror` for SeQant, if built as a top-level project by default. Controllable via `SEQUANT_WARNINGS_AS_ERRORS` CMake option
- Removes benchmarking from unit tests. Instead, benchmarks are transferred to `benchmarks/` and unit tests only test. This is warning related because the use of the profiling macro mostly relied on non-standard compiler extensions to provide the code inline (using extra encapsulating parentheses), which rightfully triggers a warning.
- Introduces a new dependency on https://github.com/nemtrif/utfcpp in order to be able to do UTF16<->UTF8 conversions without relying on deprecated `codecvt` functionality.


Krzmbrzl:
@evaleev Can we please keep the asserts that you removed in bcb8674487ecb60347970116f6bc5e941655bf01? I'm fine with using `abort` instead of throwing but `assert` gives you a location when it is triggered, which is incredibly valuable. `abort` doesn't give you anything, forcing you to re-run everything through a debugger.

PR #365, opened 2025-08-25
Revert "build executables for tests as needed"

Krzmbrzl:
This reverts commit 9e7097c1007710d64fc5c2158eba9f87ea2a4052.


evaleev:
Robert, one issue is that there are many targets like `check`/`check-sequant` that are broken by having an implicit dependence on `all`. Another (minor/annoying) is that `all` target is not for example accessible from IDEs directly. Most important that `all` is the second most likely to be broken target in deps (`install` is the first), so it's to be avoided at all costs. Building a workflow around a functional `all` target through the tree simply does not work/scale.

evaleev (approved):
I foresee that build granularity of "the entire project" even for running a subset of tests will become a problem but fine to deal with this later

PR #374, opened 2025-09-01
Detailed CC Profiling Options

ajay-mk:
`srcc` can now profile CC performance by switching Operator-level screening and topological optimizations in WickTheorem

### Changes
- Propagates `use_topology` logic correctly to `vac_av` functions
- `CC` class constructor can now take `use_topology` and `screen` as arguments. Default is true. 
- Removes unused options in `srcc`
- Updates class references in CC docs page
- Several dox fixes


Krzmbrzl:
Maybe it would make sense to extend the corresponding benchmark to screen through all combinations of those options?

ajay-mk:
> Maybe it would make sense to extend the corresponding benchmark to screen through all combinations of those options?

If `use_topology` is set to `false`, I can’t go beyond triples within a reasonable time on my laptop—CCSDTQ takes ~475 seconds on a single-thread run and more than ~3000 seconds if both screening and topology are off. I think this is too slow for benchmarks. One option would be to restrict to just doubles, but that is probably not very useful as a benchmark.

PR #376, opened 2025-09-02
`IndexSpaceRegistry` cleanup + paper example

evaleev:
added ISR use example from the paper as a unit test

### DIFF
--- a/SeQuant/core/index_space_registry.cpp
+++ b/SeQuant/core/index_space_registry.cpp
@@ -14,6 +14,16 @@ bitset_t IndexSpaceRegistry::physical_particle_attribute_mask() const {
   return physical_particle_attribute_mask_;
 }
 
+IndexSpace::QuantumNumbers IndexSpaceRegistry::physical_particle_attributes(
+    IndexSpace::QuantumNumbers qn) const {
+  return static_cast<bitset_t>(qn) & physical_particle_attribute_mask_;
+}
+
+IndexSpace::QuantumNumbers IndexSpaceRegistry::other_attributes(
+    IndexSpace::QuantumNumbers qn) const {
+  return static_cast<bitset_t>(qn) & ~physical_particle_attribute_mask_;
+}
+
 IndexSpaceRegistry IndexSpaceRegistry::clone() const {
   IndexSpaceRegistry result(*this);
   result.spaces_ =
--- a/SeQuant/core/index_space_registry.cpp
+++ b/SeQuant/core/index_space_registry.cpp
@@ -16,12 +16,12 @@ bitset_t IndexSpaceRegistry::physical_particle_attribute_mask() const {
 
 IndexSpace::QuantumNumbers IndexSpaceRegistry::physical_particle_attributes(
     IndexSpace::QuantumNumbers qn) const {
-  return static_cast<bitset_t>(qn) & physical_particle_attribute_mask_;
+  return to_bitset(qn) & physical_particle_attribute_mask_;
 }
 
 IndexSpace::QuantumNumbers IndexSpaceRegistry::other_attributes(
     IndexSpace::QuantumNumbers qn) const {
-  return static_cast<bitset_t>(qn) & ~physical_particle_attribute_mask_;
+  return to_bitset(qn) & ~physical_particle_attribute_mask_;
 }
 
 IndexSpaceRegistry IndexSpaceRegistry::clone() const {
--- a/SeQuant/core/index_space_registry.hpp
+++ b/SeQuant/core/index_space_registry.hpp
@@ -5,6 +5,7 @@
 #ifndef SEQUANT_INDEX_SPACE_REGISTRY_HPP
 #define SEQUANT_INDEX_SPACE_REGISTRY_HPP
 
+#include <SeQuant/core/bitset.hpp>
 #include <SeQuant/core/space.hpp>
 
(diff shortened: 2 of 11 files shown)


SeQuant/core/index_space_registry.cpp
```diff
@@ -14,6 +14,16 @@ bitset_t IndexSpaceRegistry::physical_particle_attribute_mask() const {
   return physical_particle_attribute_mask_;
 }
 
+IndexSpace::QuantumNumbers IndexSpaceRegistry::physical_particle_attributes(
+    IndexSpace::QuantumNumbers qn) const {
+  return static_cast<bitset_t>(qn) & physical_particle_attribute_mask_;
```
Krzmbrzl, line 19:
```suggestion
  return to_bitset(qn) & physical_particle_attribute_mask_;
```

SeQuant/core/index_space_registry.cpp
```diff
@@ -14,6 +14,16 @@ bitset_t IndexSpaceRegistry::physical_particle_attribute_mask() const {
...
+}
+
+IndexSpace::QuantumNumbers IndexSpaceRegistry::other_attributes(
+    IndexSpace::QuantumNumbers qn) const {
+  return static_cast<bitset_t>(qn) & ~physical_particle_attribute_mask_;
```
Krzmbrzl, line 24:
```suggestion
  return to_bitset(qn) & ~physical_particle_attribute_mask_;
```

SeQuant/core/index_space_registry.hpp
```diff
@@ -574,6 +575,25 @@ class IndexSpaceRegistry {
...
+  /// `valid_intersection(retrieve(space1_key),retrieve(space2_key))` is
+  /// registered
+  template <typename S1, typename S2,
+            typename = meta::EnableIfAllBasicStringConvertible<S1, S2>>
+  bool valid_intersection(S1&& space1_key, S2&& space2_key) const {
+    if (!contains(space1_key) || !contains(space2_key))
+      throw std::invalid_argument(
+          "IndexSpaceRegistry::valid_intersection(s1,s2): s1 and s2 must both "
+          "be "
+          "registered");
```
Krzmbrzl, line 591:
```suggestion
          "be registered");
```

SeQuant/core/index_space_registry.hpp
```diff
@@ -809,29 +877,79 @@ class IndexSpaceRegistry {
...
   /// @note again, this only makes sense to ask if in a SingleProduct vacuum
   /// context.
   bool is_pure_unoccupied(const IndexSpace& IS) const {
     if (!IS) {
       return false;
     } else {
-      return !IS.type().intersection(vacuum_occupied_space(IS.qns()).type());
+      return !IS.type().intersection(
+          vacuum_occupied_space(physical_particle_attributes(IS.qns())).type());
     }
```
Krzmbrzl, line 901:
```suggestion
return IS && return !IS.type().intersection(vacuum_occupied_space(physical_particle_attributes(IS.qns())).type());
```

SeQuant/domain/mbpt/space_qns.hpp
```diff
@@ -49,11 +49,13 @@ struct mask<Spin> {
...
-/// \note sLCAO basis traits use 3rd and 4th rightmost bits
+/// \note LCAO basis traits use 3rd and 4th rightmost bits
 enum class LCAOQNS : bitset_t {
-  ao = 0b000100,
-  pao = 0b001000,  // projected AO space denotes unoccupied spaces made from AO
-                   // basis and orthogonal to occupied orbitals
+  ao = 0b000100 | bitset_t(Spin::any),
+  pao =
+      0b001000 |
+      bitset_t(Spin::any),  // projected AO space denotes unoccupied spaces made
```
Krzmbrzl, line 57: Why are quantum numbers mixed like this? It seems like now it is impossible to set LCAO QNs without also setting spin QNs to `Spin::any` 🤔

tests/unit/test_space.cpp
```diff
@@ -232,4 +232,50 @@ TEST_CASE("index_space", "[elements]") {
...
+        .add_union("p", {"I", "A"}, is_complete);
+
+    // spaces whose QNs contain spin_any contain physical particles
+    isr.physical_particle_attribute_mask(spin_any);
+
+    // projected atomic orbitals = localized form of virtuals
+    const QuantumNumbersAttr lcao = 0b0100;
+    isr.add("μ̃", 0b100, lcao | spin_any);
+    // nonparticle mode: density fitting, or Cholesky, or Hubbard-Stratonivich
+    isr.add("x", 0b001, QuantumNumbersAttr{0b1000});
```
Krzmbrzl, line 259: Why not use the LCAO QN enum?

Krzmbrzl (commented):
Given that `IndexSpaceRegistry::retrieve` already throws if a given index space is not registered, I wonder whether there really is a benefit to having the explicit checks for whether the given index spaces are registered in all the newly introduced functions.

This causes us to check for existence twice for every index space and I'm not sure whether the function-specific error messages are more helpful than the error message from `retrieve(…)`. In fact, the latter might be more useful as it includes the index space label or type attributes.

tests/unit/test_space.cpp
evaleev: in the paper example I tried to avoid using existing enums to keep the example as lean as possble

PR #377, opened 2025-09-04
better support for noncovariant expressions

evaleev:


tests/unit/test_wick.cpp
```diff
@@ -250,6 +250,29 @@ TEST_CASE("wick", "[algorithms][wick]") {
...
+    {
+      // sequence of individual ops
+      const auto ops = {fann("i_1"), fcre("i_2"), fcre("i_3"),
+                        fann("i_4"), fann("i_5"), fcre("i_1")};
+      REQUIRE_NOTHROW(FWickTheorem(ops));
+      FWickTheorem w(ops);
+      REQUIRE_NOTHROW(w.full_contractions(false).compute());
+      auto partial_contractions =
+          FWickTheorem{ops}.full_contractions(false).compute();
+      REQUIRE(to_latex(partial_contractions) ==
```
Krzmbrzl, line 263: Maybe `REQUIRE_THAT` with the `EquivalentTo` matcher should be preferred to make unit test independent of (canonical) representation.

evaleev: Is there a way to use `EquivalentTo` with parsed expressions that contain `NormalOperators`?

PR #379, opened 2025-09-08
fix TN sum canonicalization

evaleev:
As of https://github.com/ValeevGroup/SeQuant/commit/af5e89dfe02ea764c16d03f8e05dd656f771a2ff `TNV{2,3}` by default ignore named index labels (older code did not ignore the labels). This makes the result of individual tensor network canonicalization independent of named index labels, hence invariant wrt relabeling. But in context of sums named index labels are meaningful (to be able to have consistent naming of external indices in different terms). User needs to be able to control this in general. The default is to ignore the labels, but the sum canonicalization will override this default.

### DIFF
--- a/tests/integration/eval/btas/scf_btas.hpp
+++ b/tests/integration/eval/btas/scf_btas.hpp
@@ -109,7 +109,8 @@ class SequantEvalScfBTAS final : public SequantEvalScf {
 
     for (auto&& [r, n] : zip(rs, nodes_)) {
       assert(!n.empty());
-      auto const& target_indices = ranges::front(n)->annot();
+      // update_amplitudes assumes that the residuals are in specific layout
+      auto const target_indices = sorted_annot(ranges::front(n)->as_tensor());
       r = funcs[st][log](n, target_indices, data_world_, cman_)
               ->template get<Tensor_t>();
     }
--- a/tests/integration/eval/ta/scf_ta.hpp
+++ b/tests/integration/eval/ta/scf_ta.hpp
@@ -96,7 +96,10 @@ class SequantEvalScfTA final : public SequantEvalScf {
 
     for (auto&& [r, n] : zip(rs, nodes_)) {
       assert(!n.empty());
-      auto const& target_indices = ranges::front(n)->annot();
+      auto trank = ranges::front(n)->as_tensor().bra_rank();
+      assert(trank == 1 || trank == 2);
+      // update_amplitudes assumes that the residuals are in specific layout
+      std::string target_indices = trank == 1 ? "a_1,i_1" : "a_1,a_2,i_1,i_2";
       r = funcs[st][log](n, target_indices, data_world_, cman_)
               ->template get<Tensor_t>();
     }


SeQuant/core/expressions/expr.cpp
```diff
@@ -166,12 +166,15 @@ bool Product::is_commutative() const {
...
 ExprPtr Product::canonicalize_impl(CanonicalizeOptions opts) {
-  // recursively canonicalize subfactors ...
+  // recursively canonicalize non-tensor subfactors (tensors will be
+  // canonicalized as part of the TN built of all tensor factors of this) ...
   ranges::for_each(factors_, [this, opts](auto &factor) {
-    auto bp = factor->canonicalize(opts);
-    if (bp) {
-      assert(bp->template is<Constant>());
-      this->scalar_ *= std::static_pointer_cast<Constant>(bp)->value();
+    if (factor.template is<AbstractTensor>() == false) {
```
Krzmbrzl, line 172:
```suggestion
    if (factor.template is<AbstractTensor>()) {
        return;
    }
```
(in my experience, early returns yield code that is easier to understand as it reduces the required nesting level)

SeQuant/core/expressions/expr.cpp
```diff
@@ -369,19 +371,25 @@ ExprPtr Sum::canonicalize_impl(bool multipass, CanonicalizeOptions opts) {
...
+  opts_copy.ignore_named_index_labels = false;
+
   const auto npasses = multipass ? 3 : 1;
   for (auto pass = 0; pass != npasses; ++pass) {
     // recursively canonicalize summands ...
     // using for_each and directly access to summands
-    sequant::for_each(summands_, [pass, &opts](ExprPtr &summand) {
+    sequant::for_each(summands_, [pass, &opts_copy](ExprPtr &summand) {
       ExprPtr bp;
       const auto rapid = (pass % 2 == 0);
```
Krzmbrzl, line 385: Only indirectly related to this PR, but do we really still need 2 passes of canonicalization? 🤔

SeQuant/core/options.hpp
```diff
@@ -35,9 +35,20 @@ enum class CanonicalizationMethod {
...
   CanonicalizationMethod method = CanonicalizationMethod::Topological;
+  /// specifies named indices; by default all indices that appear only once are
+  /// deduced to be named, but this may be misleading if e.g. single
+  /// summed-over dummy index appears in an expression
   std::optional<std::initializer_list<Index>> named_indices = std::nullopt;
+  /// whether to ignore the labels of named indices. Setting
+  /// to false will cause named indices to be treated as equivalent slots, which
+  /// the result to be independent of their labels. This does not make sense in
+  /// contexts where labels are meaningful, e.g. when canonicalizing sum of
+  /// tensor networks.
```
Krzmbrzl, line 50:
```suggestion
  /// tensor networks and will hence be overwritten.
```

SeQuant/core/tensor_network_v3.cpp
```diff
@@ -474,9 +455,9 @@ TensorNetworkV3 &TensorNetworkV3::operator=(
 
 ExprPtr TensorNetworkV3::canonicalize(
     const container::vector<std::wstring> &cardinal_tensor_labels,
-    CanonicalizationMethod method, const NamedIndexSet *named_indices_ptr) {
+    CanonicalizeOptions options) {
```
Krzmbrzl, line 458:
```suggestion
    const CanonicalizeOptions &options) {
```

SeQuant/core/tensor_network_v3.hpp
```diff
@@ -252,17 +252,13 @@ class TensorNetworkV3 {
...
-  /// indices treated as named indices
+  /// @param options see CanonicalizeOptions for the meaning of this
+  /// parameter; the default is given by CanonicalizeOptions::default_options()
   /// @return byproduct of canonicalization (e.g. phase); if none, returns
   /// nullptr
   ExprPtr canonicalize(
       const container::vector<std::wstring> &cardinal_tensor_labels = {},
-      CanonicalizationMethod method = CanonicalizationMethod::Complete,
-      const NamedIndexSet *named_indices = nullptr);
+      CanonicalizeOptions options = CanonicalizeOptions::default_options());
```
Krzmbrzl, line 261:
```suggestion
      const CanonicalizeOptions &options = CanonicalizeOptions::default_options());
```

Krzmbrzl (approved):

SeQuant/core/options.hpp
```diff
@@ -32,24 +32,46 @@ enum class CanonicalizationMethod {
...
```
evaleev, line 55:
```suggestion
  /// tensor networks and will be therefore ignored.
```

PR #383, opened 2025-09-12
cleanup of `closed_shell_CC_spintrace`

ABesharat:
- naive vs efficient dispatch is made orthogonal to the choice of biorthogonalization
- user-controllable choice of biorthogonalization via argument to the top-level function
- combined all stcc variants (many unused) into one, with command-line args controlling the method

### DIFF
--- a/SeQuant/domain/mbpt/spin.cpp
+++ b/SeQuant/domain/mbpt/spin.cpp
@@ -1005,7 +1005,7 @@ ExprPtr closed_shell_spintrace(
   // antisymmetrizer (A) and fully expanding the anti-symmetric tensors to
   // non-symmetric.
   // full_expansion: it fully expands the antisymmetrizer directly (can be used
-  // for optimal epns, however it is not an optimized way).
+  // for v2 eqs, however it is not an optimized way).
   auto partially_or_fully_expand = [&full_expansion](const ExprPtr& expr) {
     auto temp = expr;
     if (has_tensor(temp, L"A")) {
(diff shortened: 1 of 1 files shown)


SeQuant/domain/mbpt/spin.cpp
```diff
@@ -1167,7 +1167,7 @@ container::svector<ResultExpr> closed_shell_spintrace(
       is_direct_full_expansion);
 }
 
-ExprPtr closed_shell_CC_spintrace_compact_set(ExprPtr const& expr) {
+ExprPtr closed_shell_CC_spintrace_optimal(ExprPtr const& expr) {
```
evaleev, line 1170: it seems we have settled on the pattern with `spintrace` appearing at the end of the function name, so let's keep it. Here's what I propose:
- rename `closed_shell_CC_spintrace` to `closed_shell_CC_v0_spintrace`
- rename `closed_shell_CC_spintrace_compact_set` to `closed_shell_CC_v1_spintrace`
- re-introduce `closed_shell_CC_spintrace` and have it dispatch to to `closed_shell_CC_v1_spintrace` by default.
- update dox. It's probably better to refer to new equations as v1  rather than compact/optimal (not sure how to prove optimality, we may discover yet better way to do things in the future ...)

evaleev (changes requested):

SeQuant/domain/mbpt/spin.cpp
```diff
@@ -1005,7 +1005,7 @@ ExprPtr closed_shell_spintrace(
   // antisymmetrizer (A) and fully expanding the anti-symmetric tensors to
   // non-symmetric.
   // full_expansion: it fully expands the antisymmetrizer directly (can be used
-  // for compact-set epns, however it is not an optimized way).
+  // for optimal epns, however it is not an optimized way).
```
ajay-mk, line 1008:
```suggestion
  // for optimal eqns, however it is not an optimized way).
```

PR #386, opened 2025-09-16
Fix dependencies

Krzmbrzl:


cmake/modules/FindOrFetchGoogleBenchmark.cmake
```diff
@@ -10,7 +10,7 @@ if (NOT TARGET benchmark::benchmark)
         GIT_SHALLOW
         EXCLUDE_FROM_ALL
         SYSTEM
-        FIND_PACKAGE_ARGS NAMES benchmark
+        FIND_PACKAGE_ARGS 1.9.3 NAMES benchmark
```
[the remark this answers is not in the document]
evaleev: Probably should be moved to `versions.cmake` also ... To keep all prerequisite versions in one place

evaleev:
@Krzmbrzl have a look

cmake/modules/FindOrFetchSpdlog.cmake
```diff
@@ -10,7 +10,10 @@ if (NOT TARGET spdlog::spdlog)
         GIT_SHALLOW
         EXCLUDE_FROM_ALL
         SYSTEM
-        FIND_PACKAGE_ARGS NAMES spdlog
+		# The CMake option is not honored if spdlog is found via CMake config file
+		# However, our code doesn't compile when using libfmt as we're specializing templates for std::format
+		# see https://github.com/gabime/spdlog/issues/3467
+		#FIND_PACKAGE_ARGS NAMES spdlog
```
evaleev, line 16: @Krzmbrzl sounds like we need to do a 3-step dance:
- if TARGET spdlog::spdlog already available, still introspect it to detect if using std::format. If not, instist on FC.
- if TARGET spdlog::spdlog is not yet available, find_package(spdlog), introspect, if not using std::format insist on FC
- FC if needed.

external/versions.cmake
```diff
@@ -4,28 +4,22 @@
...
 
 set(SEQUANT_TRACKED_POLYMORPHICVARIANT_TAG 010c69786104c07c5faccffe0e99f99de5a69fd8)
-set(SEQUANT_TRACKED_POLYMORPHICVARIANT_PREVIOUS_TAG 010c69786104c07c5faccffe0e99f99de5a69fd8)
 
 set(SEQUANT_TRACKED_UTFCPP_TAG v4.0.6)
-set(SEQUANT_TRACKED_UTFCPP_PREVIOUS_TAG v4.0.6)
 
 set(SEQUANT_TRACKED_CLI11_TAG v2.5.0)
-set(SEQUANT_TRACKED_CLI11_PREVIOUS_TAG v2.5.0)
+set(SEQUANT_OLDEST_CLI11_VERSION 2)
```
Krzmbrzl, line 17: I think I'd vote for the term "minimum version" rather than "oldest version" :thinking:

cmake/modules/FindOrFetchSpdlog.cmake
Krzmbrzl: Introspecting seems somewhat fragile. I think I tend to stick with always using FetchContent for now (assuming that now top-level project has defined the target otherwise - in which case FC will cause a target conflict anyway). Going forward, I would switch to a different logging library instead.

evaleev: presumably can see if `SPDLOG_USE_STD_FORMAT` is among compile definitions of the spdlog target: https://github.com/gabime/spdlog/blob/v1.x/CMakeLists.txt#L287

not a big deal now but will probably need this to be adoptable by packagers

PR #389, opened 2025-09-25
spin(trace) code is moved to `mbpt` namespace

evaleev:


ajay-mk:
Should `biorthogonalization.{hpp,cpp}` be in the `mbpt` namespace as well?

Issue #393, opened 2025-09-26
Factor out Similarity Transformation Logic


ajay-mk:
The similarity transformation logic is now tied to the `mbpt::CC` class, need to factor it out for general use cases.

PR #400, opened 2025-10-01
MBPT Utility Functions

ajay-mk:
### Changelist
- Factors out similarity transformation logic from CC class: #393 
- Introduce a general function to screen out non-contributing terms from expectation values
- Update `can_change_qns` function and factor out the logic to compute quantum number change produced by a product of operators. 
- Docs++

### DIFF
--- a/SeQuant/domain/mbpt/op.cpp
+++ b/SeQuant/domain/mbpt/op.cpp
@@ -1014,8 +1014,8 @@ ExprPtr L(nann na, ncre nc, const cre<IndexSpace>& cre_space,
 
 ExprPtr L(nₚ np, nₕ nh) { return L(nann(np), ncre(nh)); }
 
-bool can_change_qns(const ExprPtr& op_or_op_product, const qns_t target_qns,
-                    const qns_t source_qns = {}) {
+bool can_change_qns(const ExprPtr& op_or_op_product, const qns_t& target_qns,
+                    const qns_t& source_qns) {
   qns_t qns = source_qns;
   if (op_or_op_product.is<Product>()) {
     const auto& op_product = op_or_op_product.as<Product>();
--- a/SeQuant/domain/mbpt/op.cpp
+++ b/SeQuant/domain/mbpt/op.cpp
@@ -1036,27 +1036,25 @@ bool can_change_qns(const ExprPtr& op_or_op_product, const qns_t& target_qns,
 }
 
 bool raises_vacuum_up_to_rank(const ExprPtr& op_or_op_product,
-                              const unsigned long k) {
+                              unsigned long k) {
   assert(op_or_op_product.is<op_t>() || op_or_op_product.is<Product>());
 
   return can_change_qns(op_or_op_product, interval_excitation_type_qns(k));
 }
 
 bool lowers_rank_or_lower_to_vacuum(const ExprPtr& op_or_op_product,
-                                    const unsigned long k) {
+                                    unsigned long k) {
   assert(op_or_op_product.is<op_t>() || op_or_op_product.is<Product>());
   return can_change_qns(op_or_op_product, qns_t{},
                         interval_excitation_type_qns(k));
 }
 
-bool raises_vacuum_to_rank(const ExprPtr& op_or_op_product,
-                           const unsigned long k) {
+bool raises_vacuum_to_rank(const ExprPtr& op_or_op_product, unsigned long k) {
   assert(op_or_op_product.is<op_t>() || op_or_op_product.is<Product>());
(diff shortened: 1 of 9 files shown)


SeQuant/domain/mbpt/op.cpp
```diff
@@ -1036,27 +1044,25 @@ bool can_change_qns(const ExprPtr& op_or_op_product, const qns_t target_qns,
...
 bool raises_vacuum_up_to_rank(const ExprPtr& op_or_op_product,
-                              const unsigned long k) {
+                              unsigned long k) {
   assert(op_or_op_product.is<op_t>() || op_or_op_product.is<Product>());
 
   return can_change_qns(op_or_op_product, interval_excitation_type_qns(k));
 }
 
 bool lowers_rank_or_lower_to_vacuum(const ExprPtr& op_or_op_product,
-                                    const unsigned long k) {
```
evaleev, line 1046: why `const` removed here and other similar functions?

SeQuant/domain/mbpt/op.hpp
```diff
@@ -1052,51 +1052,76 @@ DEFINE_SINGLE_SIGNED_ARGUMENT_OP_VARIANT(A);
...
-/// @param order order of perturbation
 /// @param K rank up to which the sum is to be formed
+/// @param order order of perturbation
 /// @param skip1 if true, skips single deexcitations
 /// @pre `order==1`, only first order perturbation is supported now
-ExprPtr Λ_pt(std::size_t order, std::size_t K, bool skip1 = false);
-
+ExprPtr Λ_pt(std::size_t K, std::size_t order = 1, bool skip1 = false);
+
+/// @brief computes the quantum number change effected by a given Operator or
```
evaleev, line 1089: change effected on what?

would `apply_to_vac` be better?

SeQuant/domain/mbpt/utils.hpp
```diff
@@ -0,0 +1,50 @@
...
+/// @param expr input expression
+/// @param skip_clone if true, will not clone the input expression
+/// @return return screened expression
+/// @code
+/// // example usage:
+/// auto expr1 = screen_zero_terms(expr); // screens for <0| expr |0>
+/// auto expr2 = screen_zero_terms(P(2) * expr); // screens for <P(2)| expr |0>
+/// @endcode
+/// @pre This function expects \p input to be composed of mbpt::Operators
+ExprPtr screen_zero_terms(ExprPtr expr, bool skip_clone = false);
```
evaleev, line 46: this screens zero terms from ref expectation value ... better name is `screen_vac_av` or `nonzero_vac_av`

re: `vac_av` vs `ref_av`, it seems indeed that what we currently call `vac_av` is `ref_av` which works correctly for the case of vac==ref (see https://github.com/ValeevGroup/SeQuant/blob/da940757199465981b15b13db84fb037a0743d9a/SeQuant/domain/mbpt/op.cpp#L1078). I propose we rename it to `ref_av` and then provide `vac_av` that computes vacuum average (i.e. for the ref!=vac case forces full contractions). The latter can be used for single reference contexts and has the expected terminology. But `ref_av` can be used in more general situations. Clearly both should just use same implementation.

SeQuant/domain/mbpt/op.hpp
ajay-mk: It was supposed to be the change effected on the vacuum. `apply_to_vac` sounds better, will update.

SeQuant/domain/mbpt/utils.hpp
```diff
@@ -0,0 +1,50 @@
...
+#include <SeQuant/core/rational.hpp>
+#include <SeQuant/core/runtime.hpp>
+
+#include <SeQuant/domain/mbpt/op.hpp>
+
+#include <range/v3/view.hpp>
+
+namespace sequant::mbpt {
+// clang-format off
+/// @brief Computes the similarity transformation e^(-B) A e^B using a commutator series truncated up to the specified order. Uses the Baker-Campbell-Hausdorff expansion.
```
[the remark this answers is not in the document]
ajay-mk: Another name to consider would be `exp_sim_tr`, not sure if that is any better. I checked some other codes to see what they do. Wick&d uses `bch` and pdaggerq refers to it as a similarity transformation.

I think it is okay to call it `sim_tr` as long as we clearly explain what it does.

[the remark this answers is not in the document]
evaleev: I agree, `sim_tr` is OK. BCH is problematic for many reasons. Campbell was the one to show the commutator expression for e^-A B e^A in 1891 (DOI 10.1112/plms/s1-28.1.381). But `campbell_transform` is too unfamiliar to users.

PR #403, opened 2025-10-02
Fix Printing of A/S/P Operators

### LINKED ISSUE #399: `to_latex()` does not include rank info for A/P operators
`to_latex()` calls on `A` or `P` operators do not display rank information.

Example: 
```c++
auto P1 = P(1);
auto P2 = P(2);
auto P_1_2 = P(nₚ(2), nₕ(1));

std::wcout << "P1 = " << to_latex(P1) << "\n";
std::wcout << "P2 = " << to_latex(P2) << "\n";
std::wcout << "P(2,1) = " << to_latex(P_1_2) << "\n";
```
All of these cases produce `{\hat{A}}`. The output should include rank information, for example: `{\hat{A}_{2}}` and `{\hat{A}_{2,1}}`

ajay-mk:
Will resolve #399

### DIFF
--- a/SeQuant/domain/mbpt/op.cpp
+++ b/SeQuant/domain/mbpt/op.cpp
@@ -354,18 +354,30 @@ std::wstring to_latex(const mbpt::Operator<mbpt::qns_t, S>& op) {
     base_lbl.pop_back();
   }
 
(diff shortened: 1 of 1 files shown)


evaleev:
need unit test(s)

PR #409, opened 2025-10-14
Custom asserts

Krzmbrzl:
This allows enabling assertions in release builds. This is quite useful as a way to make SeQuant self-validate with somewhat real-world applications (which would be way too slow in Debug builds).

### DIFF
--- a/tests/unit/test_index.cpp
+++ b/tests/unit/test_index.cpp
@@ -128,10 +128,11 @@ TEST_CASE("index", "[elements][index]") {
       REQUIRE_NOTHROW(Index{} = std::move(i7_copy));
       REQUIRE(i7_copy.nonnull() == false);
 
-#if SEQUANT_ASSERT_VARIANT == SEQUANT_ASSERT_THROW
-      REQUIRE_THROWS(Index(isr->retrieve(L"i"), 4, {i1, i1}));
-      REQUIRE_THROWS(Index(L"i_5", {L"i_1", L"i_1"}));
-      REQUIRE_THROWS(Index(L"i_5", {L"i_1", L""}));
+#if SEQUANT_ASSERT_BEHAVIOR == SEQUANT_ASSERT_THROW
+      REQUIRE_THROWS_AS(Index(isr->retrieve(L"i"), 4, {i1, i1}),
+                        sequant::Exception);
+      REQUIRE_THROWS_AS(Index(L"i_5", {L"i_1", L"i_1"}), sequant::Exception);
+      REQUIRE_THROWS_AS(Index(L"i_5", {L"i_1", L""}), sequant::Exception);
 #endif
     }
 


evaleev:
This *should* work correctly. Key is capturing the line number at the call site. Preprocessor uses #line during expansion to map line numbers in preprocessed code to line numbers in the original code. Otherwise there would be no point to source_location at all

evaleev:
> @evaleev have you verified that the line number will be correct? I didn't use `std::source_location` because I thought that the reported line number might be the one observed _after_ the macro has been expanded - I didn't test though as I could simply use the corresponding macros..

in the end there was an issue, due to a bug in clang 15 and older. Seems to work as expected.

evaleev:
FYI, I considered using Boost.Assert also ... seemed nice, but not sure if it's worth it plus not sure how this will play in a large dependency graph with multiple users of Boost.Assert ...

evaleev:
Looks like all remaining uses of `assert` are in examples (so should not be replaced by `SEQUANT_ASSERT`) and `NDEBUG` is used for non-assert purposes.

Long term we want to switch all precondition checking to SEQUANT_ASSERT; unfortunately there are still many that use `throw std::invalid_argument`. But that can be done elsewhere. At least now we have proper assert macro.

evaleev (approved):

tests/unit/test_macros.cpp
```diff
@@ -0,0 +1,46 @@
...
+      // clang-format off
+#line 1000  // to make sure the line number of the next line is fixed
+sequant::assert_failed("test");
+      // clang-format on
+    } catch (sequant::Exception& ex) {
+      // see #line up there
+      // N.B. clang <16 has std::source_location produce wrong line numbers when
+      // initialized as default argument see
+      // https://github.com/llvm/llvm-project/issues/56379
+#if defined(SEQUANT_CXX_COMPILER_IS_CLANG) && __clang_major__ >= 16
```
Krzmbrzl:
```suggestion
#if !defined(SEQUANT_CXX_COMPILER_IS_CLANG) || __clang_major__ >= 16
```
Otherwise, the test only works when using Clang, which seems undesirable

.github/workflows/cmake.yml
```diff
@@ -60,6 +60,7 @@ jobs:
         -DSEQUANT_SKIP_LONG_TESTS=${{ matrix.valgrind || matrix.sanitize }}
         -DSEQUANT_COMPILE_DOC_EXAMPLES=${{ !matrix.valgrind }}
         -DSEQUANT_PYTHON=${{ !matrix.valgrind && !matrix.sanitize }}
+        -DSEQUANT_ASSERT_BEHAVIOR=THROW
```
Krzmbrzl: I don't think I like having `throw` as the default here. This opens up the possibility of absorbing assertion failures (unintentionally) in `try…exept` blocks (and the message printed to console won't cause CI to fail)`.

I would propose overwriting the macro definition only for the macros test file and leaving it at `abort` for everything else.

SeQuant/core/utility/macros.hpp
```diff
@@ -62,13 +66,71 @@
...
+#define SEQUANT_ASSERT_BEHAVIOR \
+  SEQUANT_CONCAT(SEQUANT_ASSERT_, SEQUANT_ASSERT_BEHAVIOR_)
+#if SEQUANT_ASSERT_BEHAVIOR != SEQUANT_ASSERT_IGNORE
+#define SEQUANT_ASSERT_ENABLED
+#endif
+
+namespace sequant {
+
+inline void assert_failed(
+    [[maybe_unused]] const std::string &errmsg,
```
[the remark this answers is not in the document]
evaleev: unless `SEQUANT_ASSERT_BEHAVIOR == SEQUANT_ASSERT_IGNORE` but in that case this is broken ... will fix elsewhere

evaleev:
@Krzmbrzl addressed your points

PR #420, opened 2025-10-23
Initialize remaining members of AbstractGraph

### LINKED ISSUE #419: Uninitialized values in BLISS `AbstractGraph`
Values that are not initialized in `AbstractGraph::AbstractGraph` are:
```
AbstractGraph::failure_recording_fp_deviation
AbstractGraph::refine_current_path_certificate_index
AbstractGraph::refine_compare_certificate
AbstractGraph::refine_equal_to_first
AbstractGraph::refine_first_path_subcertificate_end
AbstractGraph::refine_cmp_to_best
AbstractGraph::refine_best_path_subcertificate_end
AbstractGraph::long_prune_max_stored_autss
AbstractGraph::long_prune_begin
AbstractGraph::long_prune_end
AbstractGraph::compute_eqref_hash
AbstractGraph::certificate_index
AbstractGraph::cr_level
AbstractGraph::cr_component_elements
```

Of those, at least `AbstractGraph::refine_equal_to_first` is/can be really used in the uninitialized state (observed via undefined behavior sanitizer).

Ideally, all members should be initialized in the constructor so they have a specific value instead of random garbage. However, the problem is that I don't really know what the appropriate initial value for all of these members are 🤔 

/CC @evaleev

Krzmbrzl:
Fixes #419

### DIFF
--- a/SeQuant/external/bliss/graph.cc
+++ b/SeQuant/external/bliss/graph.cc
@@ -56,47 +56,6 @@ namespace bliss {
  *
  *-------------------------------------------------------------------------*/
 
-AbstractGraph::AbstractGraph() {
-  /* Initialize stuff */
-  first_path_labeling = 0;
-  first_path_labeling_inv = 0;
-  best_path_labeling = 0;
-  best_path_labeling_inv = 0;
-  first_path_automorphism = 0;
-  best_path_automorphism = 0;
-  in_search = false;
-
-  /* Default value for using "long prune" */
-  opt_use_long_prune = true;
-  /* Default value for using failure recording */
-  opt_use_failure_recording = true;
-  /* Default value for using component recursion */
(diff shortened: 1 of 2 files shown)


evaleev:
@Krzmbrzl refined a bit, I think best default value is language-defined "default" (`{}`) ... let me know if this works.

Krzmbrzl:
> refined a bit, I think best default value is language-defined "default" aka [zero-initialization](https://en.cppreference.com/w/cpp/language/zero_initialization.html) ({}) ... let me know if this works.

Good point - while syntactically different, the values the fields are initialized to remain the same, so this does indeed work. Or rather, I should say: The member that I observed issues with is still explicitly initialized with your changes and the unit tests still pass, which is all I was going by anyway :D

PR #441, opened 2025-12-02
Addressing closed-shell spin-tracing issues

ABesharat:
**First issue**: closed_shell_CC_spintrace methods expect Sum only
Direct parse_expr and casting so product also work.
**Second issue**: symmetrize_expr expects the first factor to be the A tensor. So it doesn't work when ω is present.
Both symmetrize_expr and S_maps expected canonical sequence of tensors; it's solved now.

### DIFF
--- a/tests/unit/test_spin.cpp
+++ b/tests/unit/test_spin.cpp
@@ -981,16 +981,16 @@ SECTION("Closed-shell spintrace CCSD") {
 
 SECTION("Closed-shell CC spintrace for variable, constant, product") {
   {  // test variable * tensors
-    auto expr1 = sequant::parse_expr(L"-1 ω * A{i1,i2;a1,a2} * t{a1,a2;i1,i2}",
+    auto expr1 = sequant::parse_expr(L"-ω A{i1,i2;a1,a2} t{a1,a2;i1,i2}",
                                      Symmetry::Antisymm);
 
     auto result_v1 = mbpt::closed_shell_CC_spintrace_v1(expr1);
     REQUIRE_THAT(result_v1,
-                 EquivalentTo(L"-2 ω * S{i1,i2;a1,a2} * t{a1,a2;i1,i2}"));
+                 EquivalentTo(L"-2 ω S{i1,i2;a1,a2} t{a1,a2;i1,i2}"));
 
     auto result_v2 = mbpt::closed_shell_CC_spintrace_v2(expr1);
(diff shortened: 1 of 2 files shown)


tests/unit/test_spin.cpp
```diff
@@ -730,7 +730,8 @@ SECTION("Swap bra kets") {
 SECTION("Closed-shell spintrace CCD") {
   // Energy expression
   {
-    {  // standard = v1
+    {  // standard = v1 , expression can be parsed directly without requiring
+       // cast to Sum
```
ajay-mk: But the code below explicitly constructs a `Sum`. We can avoid that now.

tests/unit/test_spin.cpp
```diff
@@ -978,6 +979,14 @@ SECTION("Closed-shell spintrace CCSD") {
   }
 }  // CCSD R1
 
+SECTION("Closed-shell spintrace LR-CCSD") {
+  {  // related to issue #432
+    auto expr1 = sequant::parse_expr(L"-1 ω * A{i1,i2;a1,a2} * t{a1,a2;i1,i2}");
+    auto st_expr1 = mbpt::closed_shell_CC_spintrace_v2(expr1);
+    std::wcout << "ST Expr1: " << st_expr1.to_latex() << "\n";
```
ajay-mk: This only prints the result; need to validate it using `REQUIRE(...)`.

ajay-mk (changes requested):
Thanks, Azam. Please update the new test with actual validation of the result.

ajay-mk (approved):

evaleev (changes requested):

tests/unit/test_spin.cpp
```diff
@@ -978,6 +979,53 @@ SECTION("Closed-shell spintrace CCSD") {
   }
 }  // CCSD R1
 
+SECTION("Closed-shell CC spintrace for variable, constant, product") {
+  {  // test variable * tensors
+    auto expr1 = sequant::parse_expr(L"-1 ω * A{i1,i2;a1,a2} * t{a1,a2;i1,i2}",
```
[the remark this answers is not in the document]
ajay-mk: Yes. `L"- ω A{i1,i2;a1,a2} t{a1,a2;i1,i2}"` works. The example I provided in the issue had `-1` and `*`, this is probably copied from there.

PR #510, opened 2026-03-24
Add avoided connections support and introduce `EVOptions`

ajay-mk:
### Summary
- Add `set_nop_avoided_connections` to `WickTheorem`: forbids all contractions between specified operator pairs
- Introduce `EVOptions<T>` struct to bundle expectation value parameters (connect, avoid, screen, use_topology, skip_clone). Some of the options are only valid at Operator level. 
- Refactor `ref_av/vac_av` at both tensor and operator levels to use `EVOptions`.
- **Breaking change:**  The positional-argument signatures for `ref_av/vac_av` are removed. All call sites must use `EVOptions` or designated initializers.

```C++
// Tensor level
// before
auto result = t::vac_av(expr, {{1, 2}, {1, 3}});
auto result = t::ref_av(expr, {{0, 1}}, /* use_topology = */ false);

// now
auto result = t::vac_av(expr, {.connect = {{1, 2}, {1, 3}}});
auto result = t::ref_av(expr, {.connect = {{0, 1}}, .use_topology = false});

// Operator level
// before
auto result = vac_av(expr, op_connections, use_topology, screen);

// now
auto result = vac_av(expr, {.connect = op_connections,
                             .use_topology = use_topology,
                             .screen = screen});

// new: avoided connections
auto result = ref_av(expr, {.connect = default_op_connections(),
                             .avoid = {{antisymm_label(), antisymm_label()}}});
```
- All call sites have been updated: CC, tests, benchmarks, Python bindings
- New avoided-connections tests. 

### Finer Details
-  `OpConnections<T>` is a standalone alias and is used within `EVOptions`, `default_op_connections()`, `concat()`, `CC::ref_av` parameter etc. 
-  `EVOptions` for top-level API, separated args for impl functions. The `expectation_value_impl` functions take individual parameters rather than `EVOptions` because they recurse internally and need to flip `skip_clone = true` on recursive calls. Using EVOptions would require copying and mutating the struct on each recursive call.
-  impl declarations of expectation value methods are removed from public headers. `tensor::expectation_value_impl` is removed from `op.hpp` and forward-declared in `vac_av.cpp`. 
- `CC::ref_av `takes `OpConnections`, not `EVOptions`. The CC class owns `use_topology` and `screen` as member state and callers within CC only ever vary the connections. This ensures consistent behavior across the class.
- Python bindings expose full `EVOptions`. The `VacuumAverage` overload converts options to `EVOptions<std::wstring> `internally.

### DIFF
--- a/SeQuant/core/wick.hpp
+++ b/SeQuant/core/wick.hpp
@@ -259,8 +259,9 @@ class WickTheorem {
 
   /// @name avoided-connectivity specifiers
   ///
-  /// Ensures that the given pairs of normal operators are never directly
-  /// contracted; by default will not constrain any connectivity
+  /// Ensures that no contraction is ever made between the given pairs of
+  /// normal operators; every individual contraction attempt between an
+  /// avoided pair is rejected.
   /// @param op_index_pairs the list of pairs of op indices that must not be
   /// directly contracted
   /// @throw Exception if @p op_index_pairs contains duplicates
@@ -935,7 +936,8 @@ class WickTheorem {
 
     /// If the target connectivity will be violated by this contraction, keep
     /// the state unchanged and return false.
-    /// Also rejects contractions between avoided operator pairs.
+    /// Also rejects contractions between operator pairs specified via
+    /// set_nop_avoided_connections().
     template <typename Cursor>
     inline bool connect(const container::svector<std::bitset<max_input_size>>
                             &target_nop_connections,
--- a/SeQuant/core/wick.hpp
+++ b/SeQuant/core/wick.hpp
@@ -257,6 +257,83 @@ class WickTheorem {
   }
   ///@}
 
+  /// @name avoided-connectivity specifiers
+  ///
+  /// Ensures that the given pairs of normal operators are never directly
+  /// contracted; by default will not constrain any connectivity
+  /// @param op_index_pairs the list of pairs of op indices that must not be
+  /// directly contracted
+  /// @throw Exception if @p op_index_pairs contains duplicates
+  ///@{
+
+  /// @tparam IndexPairContainer a sequence of std::pair<Integer,Integer>
+  template <typename IndexPairContainer>
+  WickTheorem &set_nop_avoided_connections(
+      IndexPairContainer &&op_index_pairs) {
+    auto has_duplicates = [](const auto &op_index_pairs) {
+      const auto the_end = end(op_index_pairs);
+      for (auto it = begin(op_index_pairs); it != the_end; ++it) {
+        const auto found_dup_it = std::find(it + 1, the_end, *it);
+        if (found_dup_it != the_end) {
+          return true;
+        }
+      }
+      return false;
+    };
+    if (has_duplicates(op_index_pairs)) {
+      throw Exception(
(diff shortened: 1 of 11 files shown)


ajay-mk:
Reviewed locally with Claude (Opus 4.6), did not report any major issues.

evaleev (changes requested):
## Review

Well-structured PR. The `EVOptions` designated-initializer pattern is a clear improvement over the positional-argument API, and the avoided connections feature is cleanly integrated into the existing `WickTheorem` machinery. A few items to address:

### Issues

**1. Rename `avoid` → `do_not_connect`**

`avoid` is vague — it could mean "avoid evaluating" or something else. Naming it to mirror `connect` makes the semantics immediately clear:

```cpp
auto result = ref_av(expr, {.connect = {{L"f", L"t"}},
                            .do_not_connect = {{antisymm_label(), L"f"}}});
```

`disconnect` could be misread as "break an existing connection" rather than "forbid." `do_not_connect` is unambiguous and reads naturally alongside `.connect`.

**2. Potential typo in test expectation (test_mbpt.cpp)**

```cpp
const std::wstring expected4 =
    L"... t{a_3,a_2;i3,i_1}:A-C-S";
//                  ^^^ missing underscore — should be i_3?
```

Worth verifying this string is correct.

**3. Guard condition in `set_nop_avoided_connections` (wick.hpp)**

```cpp
if (expr_input_ == nullptr || !nop_avoided_connections_input_.empty()) {
```

This mirrors `set_nop_connections`, but the two branches ("input already resolved" vs. "cache for later") are easy to misread. A brief comment clarifying the intent would help.

**4. `CC::ref_av` doesn't forward `avoid`/`do_not_connect`**

The `.avoid` field is never populated in `CC::ref_av`. The PR description says "no use case right now," which is fine, but consider accepting a second parameter `OpConnections<std::wstring> do_not_connect = {}` now to avoid a second breaking change later.

**5. `EVOptions` default `connect` differs between tensor and operator levels**

- `tensor::ref_av` / `tensor::vac_av` default to `EVOptions<int>{}` → empty connections
- `op::ref_av` / `op::vac_av` default to `EVOptions<std::wstring>{.connect = default_op_connections()}`

This is correct (tensor-level doesn't know about labels), but the asymmetry means `EVOptions` defaults (`connect = {}`) are misleading at the operator level — the actual default is injected in the function signature. Consider documenting this clearly in `EVOptions`.

### Minor / non-blocking

- Consider squashing the final fixup commits ("remove debug printing", "fix container type", "rename argument") before merge.
- The `lower_label_pairs` helper in the anonymous namespace is a nice refactor.
- The forward declaration of `tensor::expectation_value_impl` in `vac_av.cpp` is reasonable, but if more impl functions need similar treatment later, a small internal header might be cleaner.

ajay-mk:
All issues have been addressed:

1. Renamed and updated all call sites.
2. Updated. This was still correct because the `"i_3"` and `"i3"` will produce the same `Index` object.
3. Added comment in `wick.hpp`. Also added the same comment when the connections are processed. 
4. Added `do_not_connect` which defaults to empty. 
5. Added notes to method documentation explaining the defaults. This was already the behavior in the master branch.

evaleev (approved):

PR #513, opened 2026-04-19
`Power` expression type

ajay-mk:
## Summary

- Introduces `sequant::Power`: a new expression type representing `base^exponent` where `base` is a `Constant` or `Variable`, and `exponent` is a `rational`.
- Adds support for `Power` in simplification, serialization, evaluation, and export.

### Changes
- core: Introduces `sequant::Power` and related logic (arithmetic, flatten, hashing, etc).
- IO: `serialize`/`deserialize` grammar: `base^(exp)`; conjugated form `(base^(exp))^*`.
- eval: `EvalExpr` / `EvalNode` support `Power`.
- mbpt: scalar-leaf filter generalized and moved from `op::expectation_value_impl` to `tensor::expectation_value_impl`.

### Examples
```c++
  // Constant base, integer exponent: 2^3
  auto p1 = ex<Power>(2, 3);

  // Variable base, rational exponent: α^(1/2)
  auto p2 = ex<Power>(L"α", rational{1, 2});

  // Simplification
  auto e = ex<Power>(ex<Constant>(2), rational{2}) * ex<Variable>(L"x");
  simplify(e);   // -> 4 * x

  // Combining like-base Powers in a product (Power::operator*=)
  Power p(L"x", 2);
  p *= Power(L"x", 3);   // x^5
  p *= Variable(L"x");   // x^6

  // Serialization grammar: base^(exp)
  auto half = deserialize<ExprPtr>(L"2^(1/2)");                  // Power(Constant(2), 1/2)
  auto x_sq = deserialize<ExprPtr>(L"x^(2)");                    // Power(Variable("x"), 2)
  auto z    = deserialize<ExprPtr>(L"(x^(1/2))^*");              // conjugated Power
  auto prod = deserialize<ExprPtr>(L"3 * x^(-1/2) * t{i1;a1}");  // Product

  // LaTeX
  Power(Constant(2), 3).to_latex();                          // "{2}^{3}"
  Power(Variable(L"x"), rational{1, 2}).to_latex();          // "{x}^{\frac{1}{2}}"
  Power(Constant(rational{1, 2}), rational{1, 2}).to_latex();// "{\frac{1}{2^{\frac{1}{2}}}}"
```

**Further examples in**: 
- `tests/unit/test_expr.cpp` (Power section) 
- `tests/unit/test_canonicalize.cpp` (Simplification and flattening)
- `tests/unit/test_parse.cpp` (Serialization grammar)
- `tests/unit/test_eval_expr.cpp` (Evaluation support)
- `tests/unit/test_eval_ta.cpp` (Evaluation of expressions involving `Power`)
- `tests/unit/test_mbpt.cpp` (Expectation value computation of expressions involving `Power`)
-  export snapshots under `tests/unit/export_tests/power_*.export_test`


Krzmbrzl:
In analogy to other compound expressions being named `Sum` and `Product`, maybe `Exponentiation` would be a more consistent name? As a non-native English speaker, I always thought `Power` isn't the technical (mathematical) term...

SeQuant/core/expressions/power.hpp
```diff
@@ -0,0 +1,199 @@
...
+      if (n % 2 != 0) value *= b;
+      n /= 2;
+      if (n > 0) b *= b;
+    }
+    if (negate) value = scalar_type{1} / value;
+
+    expr = ex<Constant>(std::move(value));
+  }
+
+  std::wstring to_latex() const override {
```
Krzmbrzl: I think adding an implementation as a free function in the `io` namespace should be preferred. Having `to_latex` as a member function only encourages users to use that, which isn't consistently available for all types.

ajay-mk: Is the plan to do this for all types? Have a `io::latex::to_string` free function and provide shorthand as `to_latex()`?

ajay-mk:
Local Claude review pointed out a few things that I need to fix, will open this up for review after that.

SeQuant/core/expressions/power.hpp
Krzmbrzl: There is no plan to remove existing member functions as that (unnecessarily) breaks backwards compstibility. However, for new types I'd just stick with the free functions.

But yeah, in principle we should already have support for all latex formattable types, unless I missed something.

SeQuant/core/export/generator.hpp
```diff
@@ -3,14 +3,62 @@
...
+#include <SeQuant/core/utility/macros.hpp>
 
+#include <sstream>
 #include <string>
 #include <string_view>
 #include <type_traits>
 
 namespace sequant {
 
+namespace detail {
```
Krzmbrzl: This shouldn't be part of `generator.hpp`. How exactly a given expression is represented (stringified) is speicifc to the actual backend that implements the generator interface. Hence, any such operations should be part of these generators. If you want to reuse this function in multiple generators, I think introducing a `util` (or similar) header that can be included would be best. Also, ideally you'd implement this as a header with only function declarations and a cpp file with the actual definition.

SeQuant/core/export/itf.hpp
```diff
@@ -400,6 +400,13 @@ class ItfGenerator : public Generator<Context> {
       return represent(expr.as<Variable>(), ctx);
     } else if (expr.is<Constant>()) {
       return represent(expr.as<Constant>(), ctx);
+    } else if (expr.is<Power>()) {
+      // Use Python like syntax for ITF
+      const Power &power = expr.as<Power>();
+      const Expr &base = *power.base();
+      return detail::format_power_base(base, stringify(base, ctx)) + "**" +
+             detail::format_power_exponent(power.exponent(),
+                                           /*double_slash*/ false);
```
Krzmbrzl: ITF only supports this as long as the base is a constant so that evaluation of the power (via Python) leads to a constant value.

This should throw an exception if the base is anything but a constant.

SeQuant/core/export/generator.hpp
Krzmbrzl, line 83: We should extend this to an overload that does this for powers and then that should be used in all generators instead of the proposed free-standing formatting functions

SeQuant/core/export/python_einsum.hpp
```diff
@@ -475,6 +475,29 @@ class PythonEinsumGeneratorBase : public Generator<Context> {
...
+        throw Exception("Power base must be a Constant or Variable");
+      }
+      return detail::format_power_base(base, stringify_scalar(base, ctx)) +
+             "**" +
+             detail::format_power_exponent(power.exponent(),
+                                           /*double_slash*/ false);
+    }
+    throw Exception(
+        "stringify_scalar: expression is not a scalar leaf "
+        "(Variable/Constant/Power)");
```
Krzmbrzl: To be as helpful as possible, the exception message should contain the type name of `expr`

SeQuant/core/expressions/power.hpp
```diff
@@ -0,0 +1,203 @@
...
+      base_ = inner.base()->clone();
+      exponent_ = inner.exponent() * exponent_;
+    } else {
+      SEQUANT_ASSERT(base->is<Constant>() || base->is<Variable>());
+      base_ = base->clone();
+    }
+    // 0^n is defined only for n >= 0 (0^0 = 1 by convention)
+    if (base_->is<Constant>() && base_->as<Constant>().is_zero()) {
+      SEQUANT_ASSERT(exponent_ >= 0);
+    }
```
Krzmbrzl:
```suggestion
    SEQUANT_ASSERT(!base_->is<Constant>() ||  base_->as<Constant>().is_zero() ||  !exponent_ >= 0);
```
that will make sure that the full condition of this assertion will be part of a potential error message

SeQuant/core/expressions/power.hpp
```diff
@@ -0,0 +1,203 @@
...
+      if (n % 2 != 0) value *= b;
+      n /= 2;
+      if (n > 0) b *= b;
+    }
+    if (negate) value = scalar_type{1} / value;
+
+    expr = ex<Constant>(std::move(value));
+  }
+
+  std::wstring to_latex() const override;
```
Krzmbrzl: Imo we should remove this member entirely 👀

SeQuant/core/expressions/power.hpp
```diff
@@ -0,0 +1,203 @@
...
+
+  ExprPtr clone() const override {
+    return ex<Power>(base_->clone(), exponent_);
+  }
+
+  /// @brief adjoint of Power: calls adjoint on base; exponent is real.
+  void adjoint() override {
+    base_ = ::sequant::adjoint(base_);
+    reset_hash_value();
+  }
```
Krzmbrzl: That seems like it is doing the wrong thing. For the example `i^2 = -1`, taking the adjoint doesn't change anything as the result is purely real. Taking the adjoint of a power must take the adjoint of the result based on by understanding of operator precedence. That is `p*`  for `p = a^b` evaluates to `(a^b)*`.

That means, you'd have to make `adjoint` a property of `Power` as it can only be applied immediately in case the power evaluates to a constant.

SeQuant/core/expressions/power.hpp
```diff
@@ -0,0 +1,203 @@
...
+};
+
+}  // namespace sequant
+
+namespace sequant::io::latex {
+inline std::wstring to_string(const Power& power) {
+  if (power.exponent() == 1) return power.base()->to_latex();
+  return power.base()->to_latex() + L"^" + to_string(power.exponent());
+}
+}  // namespace sequant::io::latex
```
Krzmbrzl: This should be implemented within the corresponding I/O files. That's where everything else is implemented as well.

SeQuant/core/io/serialization/v1/deserialize.cpp
```diff
@@ -116,7 +118,9 @@ auto tensor_def       = x3::lexeme[
                             name >> x3::skip[index_groups] >> -(symmetry_spec)
                         ];
 
-auto nullary          = number | tensor | variable;
+auto power_def        = (number | variable) >> x3::lit(L"^") >> '(' >> number >> ')';
```
Krzmbrzl: I would probably define the grammar such that `^` is a binary operator (with higher precedence than `*`) and then reject unsupported cases while traversing the AST. That will lead to much more comprehensible error messages.

Also, you'll want parenthesis around the base in case it is a number as well. Otherwise, you'll end up with the same ambiguity as with omitted parentheses around the exponent.

SeQuant/core/io/serialization/v1/serialize.cpp
```diff
@@ -145,6 +145,26 @@ std::wstring to_string(const Variable& variable, const SerializationOptions&) {
...
+  const ExprPtr& base = power.base();
+  if (base->is<Constant>()) {
+    serialized += to_string(base->as<Constant>(), options);
+  } else if (base->is<Variable>()) {
+    serialized += to_string(base->as<Variable>(), options);
+  } else {
+    throw Exception(
+        "Cannot serialize Power with non-scalar base (expected Constant or "
+        "Variable)");
+  }
```
Krzmbrzl: Just dispatch to the generic `Expr &` overload. That seems easier to maintain than every function doing explicit type checking itself.

tests/unit/export_tests/power_fractional_base_scalar.export_test
```diff
@@ -0,0 +1,36 @@
...
+
+=================================================
+Python (einsum)
+=================================================
+import numpy as np
+import os
+
+e = 0.0
+x = np.load('x.npy')
+e += (3/2)**2 * x * np.einsum('->', optimize=True)
```
Krzmbrzl: Is that einsum call valid?

tests/unit/test_power.cpp
Krzmbrzl, line 1: Imo, these tests should simply go into `test_expr` as a `Power` is still an expression

SeQuant/core/export/generator.hpp
ajay-mk: Done.

SeQuant/core/export/itf.hpp
ajay-mk: Done.

SeQuant/core/expressions/power.hpp
ajay-mk: I am in favor of keeping this because most other types have it.

SeQuant/core/expressions/power.hpp
ajay-mk: I introduced a conjugated flag, which tracks the state. Conjugation is now only applied at the end, when Power is folded into Constant. The `i^2 = -1` case is now a unit test.

SeQuant/core/expressions/power.hpp
ajay-mk: Done.

tests/unit/export_tests/power_fractional_base_scalar.export_test
ajay-mk: This was an oversight on my end. Fixed.

SeQuant/core/expressions/power.hpp
Krzmbrzl: Many but far from all. Imo that's an inconsistent API (which we can't change for existing types dues to backwards compatibility) that I wouldn't want to extend to new types. This only encourages users do do things the wrong way until they encounter the cases where it breaks. That will be confusing to them 👀

Krzmbrzl:
Do we have a test case for exporting a conjugated power yet? If not, that would probably be worth it to add one (or more)

SeQuant/core/io/serialization/v1/serialize.cpp
```diff
@@ -145,6 +145,24 @@ std::wstring to_string(const Variable& variable, const SerializationOptions&) {
...
+                       const SerializationOptions& options) {
+  std::wstring core;
+  const auto& base = power.base();
+  // parenthesize bases for Constants and conjugated Variables
+  const bool parenthesize_base =
+      base->is<Constant>() ||
+      (base->is<Variable>() && base->as<Variable>().conjugated());
+  if (parenthesize_base) core += L"(";
+  core += v1::to_string(*base, options);
+  if (parenthesize_base) core += L")";
```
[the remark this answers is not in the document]
ajay-mk: `Power` of `Power` is not supported anymore. The base must be strictly `Constant` or `Variable`.

SeQuant/core/export/export.hpp
```diff
@@ -304,14 +324,28 @@ bool prune_scalar_factor(ExportNode<T> &node, PreprocessResult &result,
...
+    const Power &pw = factor->as<Power>();
+    if (pw.base()->is<Variable>()) {
+      if ((prunable & PrunableScalars::Variables) == PrunableScalars::None) {
+        return false;
+      }
+      result.variables[pw.base()->as<Variable>()] |=
+          node.leaf() ? Usage::Terminal : Usage::Intermediate;
+    } else if ((prunable & PrunableScalars::Constants) ==
+               PrunableScalars::None) {
+      return false;
```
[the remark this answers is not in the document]
ajay-mk: `Power` of `Power` is not supported anymore. The base must be strictly `Constant` or `Variable`.

evaleev (approved):

PR #520, opened 2026-05-20
Refactor optimization pipeline.

bimalgaudel:


SeQuant/core/optimize/optimize.hpp
```diff
@@ -9,18 +9,39 @@ namespace sequant {
...
+/// \copydoc optimize(ExprPtr const&, OptimizeOptions)
 ResultExpr& optimize(ResultExpr& expr, OptimizeOptions opts = {});
 
+/// \copydoc optimize(ExprPtr const&, OptimizeOptions)
 [[nodiscard]] ResultExpr& optimize(ResultExpr&& expr,
                                    OptimizeOptions opts = {});
 
+// Overloads for backwards compatibility
+
+/// \deprecated Use the \c OptimizeOptions overload instead.
```
Krzmbrzl, line 28: We can (and maybe should) make use of the `[[deprecated]]` attribute: https://en.cppreference.com/cpp/language/attributes/deprecated

bimalgaudel: I thought about that. I might add it.

evaleev (commented):
## Review: Refactor optimization pipeline

Solid, well-structured refactor. I did a focused thread-safety audit of the new parallel path and added minimal regression tests + documentation; details below. Pushed as `a1d0264` on this branch.

### Parallelism (`optimize_impl(..., parallel_outer=true)`) — SAFE as written, conditional
A four-angle audit (TensorCanonicalizer · TensorNetwork/bliss · Index/Expr/registry · `for_each` mechanics) found no data race, resting on two invariants that currently hold:
1. `TensorNetwork` clones every tensor on insertion, so the unsynchronized `mutable` lazy caches (`Expr::hash_value_`, `Index::label_`/`full_label_`) are only ever written on thread-local clones. `Index::operator==`/`<=>` compare immutable members only, so building index sets over shared indices is safe.
2. The one step that lazily mutates shared `Index`/`Expr` caches — `binarize()` via `Index::label()` — runs in the **sequential** loop *after* `for_each` joins (`optimize.cpp`), never inside `do_term`.

Canonicalizer/bliss global state is read-only during canonicalization; the swap counter is `thread_local`; the loop writes disjoint, pre-sized `new_smands[i]` slots.

I documented invariants (1)/(2) in a comment at the parallel branch so a future edit doesn't silently break them by moving a `->hash_value()`/`Index::label()` call into `do_term`.

**Caveats worth knowing:**
- The default `Context` and `cardinal_tensor_labels` must be configured *before* entering `optimize()` (their writes are unsynchronized unless `SEQUANT_CONTEXT_MANIPULATION_THREADSAFE`); `optimize()` only reads them.
- On Linux/GCC builds `for_each` uses `std::execution::par_unseq`, which **ignores `num_threads()`** — so `set_num_threads(1)` will not serialize this loop there (it does on the Apple-Clang/macOS build, which uses the manual-thread fallback).

### Tests added (commit `a1d0264`)
- `OptimizeOptions` API: both `OptFor::Flops` and `OptFor::Memsize` binarize a product correctly; `ReorderSum` knob; bare `optimize()` == explicit `Reorder`.
- Parallel/sequential equivalence: a 3-summand sum optimized under `set_num_threads(1)` vs `(4)` must produce identical results — the safety net for the invariants above.

### Minor doc comments added
- `flops_counter` vs `memsize_counter`: documented why flops needs `<IndexSet>` (concatenated operands repeat contracted indices) while memsize can use the default vector (per-operand, no dups).
- Documented the exact-equality `== 1.`/`!= 1.` scalar sentinel convention in both counters.

### Strengths (unchanged from the PR)
- Placeholder-label hardening (`@__opt_` + digit-by-digit parse with asserts) fixes the old `I_`-prefix `std::stoi`-on-raw-`data()` collision risk.
- `reorder` no longer re-binarizes (precomputed nodes threaded through, with `SEQUANT_ASSERT(nodes.size() == expr.size())`).
- Clean extraction of `init_results`/`build_subnet_metadata` (correctly `inline`); deprecated bool overloads correctly drop their default args to avoid ambiguity with the `OptimizeOptions = {}` overload.

LGTM once CI is green.

PR #537, opened 2026-06-01
 Support open-shell CC spintrace of energy like expressions

ajay-mk:
Needed for the unified processing and evaluation pipeline in MPQC4. The closed-shell entry point already supports energy like expressions, and this PR restores the API symmetry.

### DIFF
--- a/SeQuant/domain/mbpt/spin.cpp
+++ b/SeQuant/domain/mbpt/spin.cpp
@@ -1590,7 +1590,7 @@ std::vector<ExprPtr> open_shell_CC_spintrace(const ExprPtr& expr) {
   };
 
   // Expressions without a leading antisymmetrizer (e.g. CC energy) collapse
-  // to a single spin case. They are spin-traced directly, but term-by-term.
+  // to a single expression. They are spin-traced directly, but term-by-term.
   const auto A_opt = leading_antisymmetrizer(expr);
   if (!A_opt) {
     auto trace_term = [](const ExprPtr& term) -> ExprPtr {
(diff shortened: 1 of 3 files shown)


SeQuant/domain/mbpt/spin.hpp
```diff
@@ -366,10 +366,13 @@ std::vector<ExprPtr> open_shell_spintrace(
 /// the expectation value is with respect to an open-shell Fermi vacuum.
 /// Antisymmetrizer is expanded partially to produce antisymmetrizer for
 /// spin-up and spin-down columns.
+/// Expressions without a leading antisymmetrizer (e.g. a CC energy expression)
+/// collapse to a single spin case and the returned vector then contains one expression.
```
ABesharat: Adding a clarifying line would help:
"That expression is itself the sum over all internal spin combinations."

ABesharat (approved):

SeQuant/domain/mbpt/spin.cpp
```diff
@@ -1576,9 +1576,36 @@ std::vector<ExprPtr> open_shell_spintrace(
 
 std::vector<ExprPtr> open_shell_CC_spintrace(const ExprPtr& expr) {
   SEQUANT_ASSERT(expr->is<Sum>() || expr->is<Product>());
-  Tensor A = expr.is<Sum>() ? expr->at(0)->at(0)->as<Tensor>()
-                            : expr->at(0)->as<Tensor>();
-  SEQUANT_ASSERT(A.label() == reserved::antisymm_label());
+  // Returns the leading antisymmetrizer of an expression, or nullopt
+  // when there is none
+  auto leading_antisymmetrizer = [](const ExprPtr& e) -> std::optional<Tensor> {
```
Krzmbrzl: Don't we already have a function for that somewhere? If not, we should create one and use it everywhere where we need this functionality to not reinvent the wheel everytime.

Krzmbrzl (approved):

SeQuant/domain/mbpt/spin.cpp
ajay-mk: Switched over to `pop_tensor`.

evaleev (approved):

PR #538, opened 2026-06-02
Generalize `AsyCost`

### LINKED ISSUE #306: `AsyCost` doesn't work for generalized index spaces
As can be seen in
https://github.com/ValeevGroup/SeQuant/blob/44ce3e5a4b6293884655d68ca8e1f5b3fc0f8288/SeQuant/core/asy_cost.hpp#L13-L28
the `AsyCost` class is hardcoded to work for exactly two index spaces that can be categorized into `occupied` and `virtual`.

This model breaks down in multireference theories where expressions will involve at least a third index space (`active`). And for more general cases (e.g. mixing of fermionic and bosonic vector spaces) this will obviously fail as well.

To be more general, the class should eventually be refactored to use a `map<IndexSpace, std::size_t>` instead of explicit member variables for every space.

ajay-mk:
## Generalize `AsyCost` from O/V to arbitrary `IndexSpace`

`AsyCost` previously hard-coded asymptotic cost as a polynomial in exactly two orbital spaces: occupied (O) and virtual (V), with each entry storing a fixed (occ, virt) exponent pair. 

This PR replaces the fixed O/V representation with a general map from `IndexSpace` to integer exponent, so a cost term is now a rational multiplier times a product of arbitrary space sizes raised to integer powers

### Changes
- A cost entry now stores a map from `IndexSpace` to exponent instead of fixed occupied/virtual powers, so any number of spaces is supported.
- Constructors take an exponent map; `ops()` takes a map of space extents instead of two numbers.
- Printing uses base labels of each `IndexSpace`. 
- Cost ordering now compares total scaling first, then breaks ties space by space. 
- `test_asy_cost.cpp` and `test_eval_node.cpp` updated to construct costs via the exponent-map API, and new tests are added with general `IndexSpace`s.

### DIFF
--- a/SeQuant/core/asy_cost.cpp
+++ b/SeQuant/core/asy_cost.cpp
@@ -129,15 +129,28 @@ bool AsyCost::AsyCostEntry::operator<(const AsyCost::AsyCostEntry &rhs) const {
   // The max() sentinel is the greatest entry.
   if (is_max_ != rhs.is_max_) return !is_max_;
 
-  // Order by exponents, space by space, from the greatest IndexSpace (highest
-  // priority) down to the least, using IndexSpace's own ordering. The first
-  // space whose exponents differ decides: a larger exponent on a higher-
-  // priority space makes the entry larger. Entries with identical exponents
-  // on every space are equivalent.
+  // Order primarily by the sum of all exponents, i.e. the overall polynomial
+  // degree (worst-case scaling). When two entries share the same total degree,
+  // break the tie by comparing exponents space by space, from the greatest
+  // IndexSpace (highest priority) down to the least, using IndexSpace's own
+  // ordering: the first space whose exponents differ decides, and a larger
+  // exponent on a higher-priority space makes the entry larger. Entries with
+  // identical exponents on every space are equivalent.
   auto exp_for = [](ExponentMap const &m, IndexSpace const &s) {
     auto it = m.find(s);
     return it == m.end() ? std::size_t{0} : it->second;
   };
+
+  auto total_degree = [](ExponentMap const &m) {
+    std::size_t sum = 0;
+    for (auto const &kv : m) sum += kv.second;
+    return sum;
+  };
+  {
+    auto const l = total_degree(exponents_);
+    auto const r = total_degree(rhs.exponents_);
+    if (l != r) return l < r;
+  }
 
   container::set<IndexSpace> spaces;
   for (auto const &kv : exponents_) spaces.insert(kv.first);
--- a/SeQuant/core/asy_cost.cpp
+++ b/SeQuant/core/asy_cost.cpp
@@ -2,7 +2,6 @@
 #include <SeQuant/core/container.hpp>
 #include <SeQuant/core/meta.hpp>
 #include <SeQuant/core/rational.hpp>
-#include <SeQuant/core/utility/exception.hpp>
 #include <SeQuant/core/utility/string.hpp>
 
 #include <boost/numeric/conversion/cast.hpp>
@@ -190,11 +189,9 @@ double AsyCost::ops(ExtentMap const &extents) const {
     double temp = 1;
     for (auto const &[space, exp] : c.exponents()) {
       auto it = extents.find(space);
-      if (it == extents.end())
-        throw Exception("AsyCost::ops: no extent provided for index space '" +
-                        toUtf8(space.base_key()) + "'");
-      temp *=
-          std::pow(static_cast<double>(it->second), static_cast<double>(exp));
+      auto const extent =
+          it != extents.end() ? it->second : space.approximate_size();
+      temp *= std::pow(static_cast<double>(extent), static_cast<double>(exp));
     }
     total += boost::numeric_cast<double>(c.prefactor()) * temp;
   }
--- a/SeQuant/core/asy_cost.cpp
+++ b/SeQuant/core/asy_cost.cpp
@@ -18,26 +18,26 @@
 namespace sequant {
 
 AsyCost::AsyCostEntry::AsyCostEntry()
-    : exponents_{}, count_{0}, is_max_{false} {}
+    : exponents_{}, prefactor_{0}, is_max_{false} {}
 
-AsyCost::AsyCostEntry::AsyCostEntry(ExponentMap exponents, rational count)
-    : exponents_{std::move(exponents)}, count_{count}, is_max_{false} {
+AsyCost::AsyCostEntry::AsyCostEntry(ExponentMap exponents, rational prefactor)
+    : exponents_{std::move(exponents)}, prefactor_{prefactor}, is_max_{false} {
   for (auto it = exponents_.begin(); it != exponents_.end();) {
     if (it->second == 0)
       it = exponents_.erase(it);
     else
       ++it;
   }
-  if (count_ == 0 || exponents_.empty()) {
+  if (prefactor_ == 0 || exponents_.empty()) {
     exponents_.clear();
-    count_ = 0;
+    prefactor_ = 0;
   }
 }
 
 AsyCost::AsyCostEntry AsyCost::AsyCostEntry::max() {
   AsyCostEntry e;
   e.is_max_ = true;
-  e.count_ = std::numeric_limits<intmax_t>::max();
+  e.prefactor_ = std::numeric_limits<intmax_t>::max();
   return e;
 }
 
@@ -50,12 +50,12 @@ AsyCost::ExponentMap const &AsyCost::AsyCostEntry::exponents() const {
   return exponents_;
 }
 
-rational AsyCost::AsyCostEntry::count() const { return count_; }
+rational AsyCost::AsyCostEntry::prefactor() const { return prefactor_; }
 
-void AsyCost::AsyCostEntry::set_count(rational n) const { count_ = n; }
+void AsyCost::AsyCostEntry::set_prefactor(rational n) const { prefactor_ = n; }
 
 bool AsyCost::AsyCostEntry::is_zero() const {
-  return !is_max_ && exponents_.empty() && count_ == 0;
+  return !is_max_ && exponents_.empty() && prefactor_ == 0;
 }
 
 bool AsyCost::AsyCostEntry::is_max() const { return is_max_; }
@@ -78,8 +78,8 @@ std::string AsyCost::AsyCostEntry::text() const {
   } else if (is_zero()) {
     oss << "zero";
   } else {
-    auto abs_c = abs(count_);
-    oss << (count_ < abs_c ? "- " : "");
+    auto abs_c = abs(prefactor_);
+    oss << (prefactor_ < abs_c ? "- " : "");
     if (abs_c == 1) {
       // do nothing
     } else {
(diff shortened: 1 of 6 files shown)


SeQuant/core/asy_cost.cpp
```diff
@@ -98,63 +114,83 @@ std::string AsyCost::AsyCostEntry::to_latex() const {
...
 bool AsyCost::AsyCostEntry::operator<(const AsyCost::AsyCostEntry &rhs) const {
-  return virt() < rhs.virt() || (virt() == rhs.virt() && occ() < rhs.occ());
+  // The max() sentinel is the greatest entry.
+  if (is_max_ != rhs.is_max_) return !is_max_;
+
+  // Order by exponents, space by space, from the greatest IndexSpace (highest
+  // priority) down to the least, using IndexSpace's own ordering. The first
+  // space whose exponents differ decides: a larger exponent on a higher-
+  // priority space makes the entry larger. Entries with identical exponents
+  // on every space are equivalent.
```
Krzmbrzl: Sorting by the sum of all exponents first could be a useful criterion as oftentimes one is interested in the overall (worst-case) scaling of a method.

SeQuant/core/eval/eval_node.hpp
```diff
@@ -36,19 +37,19 @@ namespace {
...
-    return idx.space() ==
-           get_default_context().index_space_registry()->hole_space(
-               idx.space().qns());
-  });
-  auto nvirt = bk_rank - nocc;
-  return {nocc, nvirt};
+/// Tally indices per IndexSpace appearing in the bra+ket of `t`. AsyCost is
+/// only used for analysis/reporting, so spaces are taken verbatim from the
+/// indices, with no registry lookup.
+[[maybe_unused]] AsyCost::ExponentMap space_counts(Tensor const& t) {
```
Krzmbrzl: What's the `[[maybe_unused]` meant to imply here? To me it seems like an odd place to put this attribute 👀

ajay-mk: It was a suggestion from the IDE and I didn't really pay attention to it. Removing it actually breaks the build. `space_counts` lived in an anonymous namespace, it has internal linkage, every TU which includes the header and never called warns. 
I think the correct thing to do is to move these functions into a detail namespace ([9c56ce5](https://github.com/ValeevGroup/SeQuant/pull/538/commits/9c56ce50c34b3689b0f9ce9e47e047b579791e46))

SeQuant/core/asy_cost.cpp
ajay-mk: Done, and added a test case ([512041d](https://github.com/ValeevGroup/SeQuant/pull/538/commits/512041d974629a64b7e98961542b864d8360a016))

bimalgaudel (approved):

SeQuant/core/asy_cost.hpp
```diff
@@ -46,23 +54,25 @@ class AsyCost {
 
     AsyCostEntry &operator=(AsyCostEntry &&) = default;
 
-    size_t occ() const;
+    [[nodiscard]] ExponentMap const &exponents() const;
 
-    size_t virt() const;
-
-    rational count() const;
+    [[nodiscard]] rational count() const;
```
Krzmbrzl: I would suggest to use the term `prefactor` instead of `count` as that is more commonly what this is referred to in the literature (afaict at least).

SeQuant/core/asy_cost.hpp
```diff
@@ -116,10 +109,14 @@ class AsyCost {
...
-  /// \return Scaled asymptotic cost.
-  [[nodiscard]] double ops(size_t nocc, size_t nvirt) const;
+  /// Substitute each space in this cost by an extent and evaluate.
+  /// \param extents Map from index space to extent (size). Every space
+  ///                appearing in this cost must be present in `extents`.
+  /// \throw sequant::Exception if a space appearing in this cost is missing
+  ///                from `extents`.
+  /// \return Numerical value of the cost.
+  ///
+  [[nodiscard]] double ops(ExtentMap const &extents) const;
```
Krzmbrzl: Why do we pass in an extent map here instead of relying on the sizes associated with the index spaces? Iirc the latter is what the `optimize` routine does by now which would suggest doping the same here as well 🤔

tests/unit/test_asy_cost.cpp
```diff
@@ -20,70 +25,87 @@ struct MatFlops {
   }
 };
 
+// Two distinct index spaces used to build occupied/virtual costs. The type
+// bits make `virt_space` order above `occ_space`, so the virtual exponent
+// dominates cost comparison (and base_key drives how each space prints).
+sequant::IndexSpace const occ_space{L"O", 0b01};
+sequant::IndexSpace const virt_space{L"V", 0b10};
+
+sequant::AsyCost::ExponentMap ov(std::size_t nocc, std::size_t nvirt) {
```
Krzmbrzl: I'd vote for giving this function a more descriptive name and then just make it return the constructed `AsyCost` object instead of only the exponent map.

Krzmbrzl (approved):

SeQuant/core/asy_cost.hpp
bimalgaudel: This is for cost estimation/computation from the symbolic asycost object.

ajay-mk: We can substitute any extent here, independent of each space's `approximate_size()`. `optimize` also takes a configurable provider, but falls back to `approximate_size()` when nothing is supplied. 
Should I add an overload which uses the approximate sizes?

SeQuant/core/asy_cost.hpp
evaleev: In any case, needs dox. Even prefactor is not self-describing.

SeQuant/core/asy_cost.hpp
evaleev: > Should I add an overload which uses the approximate sizes?

I think natural to fallback to `approximate_size()` if space is not present in `extents`. Then default `extents` to null (empty) to fallback to `approximate_size()` for all spaces. No need for overload.

SeQuant/core/asy_cost.hpp
ajay-mk: Renamed and docs added.

SeQuant/core/asy_cost.hpp
ajay-mk: Done, and added a test.

tests/unit/test_asy_cost.cpp
ajay-mk: Done, similar change in test_eval_node also.

evaleev (commented):
## Review: Generalize `AsyCost`

### Overview

This PR replaces `AsyCost`'s hard-coded two-space (occupied `O` / virtual `V`) cost representation with a general `IndexSpace → exponent` map. A cost term becomes a rational prefactor times a product of arbitrary index-space sizes raised to integer powers. The API changes accordingly: constructors take an `ExponentMap`, `ops()` takes an `ExtentMap` of per-space sizes, printing uses each space's `base_key()`, and ordering compares total polynomial degree first, then space-by-space. `eval_node.hpp` is rewritten to tally indices per `IndexSpace` (now including `aux` slots, so DF/THC auxiliary spaces show up), and tests are migrated plus extended with multi-space, density-fitting, batched-index, and MRCC cases.

This is a clean, well-motivated generalization. Code quality is high, documentation is thorough, and test coverage is genuinely improved. Points worth discussing below.

### Correctness

**The new ordering loses the implicit V≫O weighting — please confirm this is intended.** The old `operator<` compared the virtual exponent first, then occupied. Since `V` is typically much larger than `O`, that ordering tracked true magnitude reasonably well. The new ordering sorts by **total degree** (sum of exponents) first. These disagree whenever total degrees differ:

- `O²V⁴` (degree 6) vs `O⁵V²` (degree 7): old ranks `O²V⁴` greater; new ranks `O⁵V²` greater. With `V≫O`, `O²V⁴` is the larger cost, so the new ordering disagrees with actual magnitude.

This is defensible for a space-agnostic representation (you can't weight spaces without their sizes), and the same-degree tie-break by `IndexSpace` priority reproduces the old virt-dominant behavior *within* a degree. The key mitigant — which I verified — is that `AsyCost` is **analysis/reporting only**: it is consumed solely by `Flops`/`Memory`/`min_storage`/`asy_cost` in `eval_node.hpp` and by `peak_cache` (`std::max(curr, max)` in `cache_manager.cpp`). The `optimize` module does **not** use `AsyCost` for path selection, so contraction-order decisions are unaffected. Worth a one-line note in the PR/commit that this changes how `peak_cache` and reported costs rank, even though no optimizer decision depends on it.

**`ops()` dropped the `temp > 1 ? … : 0` guard — minor behavior change.** The old code skipped a term whose substituted value was ≤ 1; the new loop always adds `prefactor * temp`. Since entries with empty exponents collapse to zero and never persist, the only divergence is when an extent of 0 or 1 is supplied (`pow(1, e) = 1`), which previously zeroed the term and now keeps it. Harmless for realistic extents (≥10) and arguably more correct — just flagging it.

**Batched-index logic is correct.** `ContractedIndexCount` now dedupes all indices across L/R/T into a `set` and counts distinct indices per space. A contracted index (in both operands) and an external index (operand + result) each collapse to one entry; a batched index (in L, R **and** T) also collapses to one, yielding the right `a³z¹` scaling. The `batched index` test confirms this. 👍

### Code Quality

- Excellent doc comments on the new API and the ordering rationale; the `space_counts` note about union-space indices (MR hole `I=i∪u` counted as a single symbol, not decomposed) is exactly the kind of subtlety worth recording.
- Moving the helpers from an anonymous namespace into `detail` and fixing the `is_outerpod` → `is_outerprod` typo are nice cleanups.
- **Dead code:** `ContractedIndexCount::rank(NodePos)` now has no consumers — `Flops` only uses `unique_counts()` and `is_outerprod()`. `ranks_` is still needed internally for the outer-product check, but the public `rank()` getter can be dropped.
- **Micro-nit:** the zero-exponent purge in the constructor could be `std::erase_if(exponents_, [](auto& kv){ return kv.second == 0; })`, though the manual loop is fine.

### Tests

Strong coverage: multi-space (>2) printing/equality/eval, extent fallback to `approximate_size()`, default-empty `ops()`, total-degree vs priority ordering, DF factorization, batched indices, and an MRCC active-space case. Two small suggestions:

- The `generalized spaces` LaTeX assertions check substring presence but not term **order**. Since ordering is a core behavior change, consider one assertion on the full rendered string to pin down term sequence.
- No test exercises `AsyCost::max()` flowing through `ops()` (should return `infinity`) or through the new `is_max` branch of `operator<`. Cheap to add and the new `is_max_` sentinel logic is otherwise only indirectly covered.

### Risk

Low. The blast radius is contained to eval cost reporting; all production call sites (`eval_node.hpp`, `cache_manager`) are updated in this PR and no external callers of the removed `AsyCost(size_t,size_t)` / `ops(size_t,size_t)` overloads remain. The main thing to consciously sign off on is the **total-degree-first ordering** as the intended new semantics for `peak_cache`/reported costs.

### Recommendation

Approve with minor follow-ups: (1) note the ordering-semantics change in the commit/PR description, (2) drop the now-unused `rank()` getter, (3) optionally add `max()`-through-`ops()` and term-order tests.

evaleev:
good to merge once checks pass

PR #542, opened 2026-06-06
core/tensor_network: braket-symmetric tensor shade invariant under bra↔ket exchange

evaleev:
## Summary

`VertexPainterImpl::to_hash_value()` computes a tensor's **shade** — the color that salts every one of its slot/index vertices in the canonicalization graph — from a hash that included `bra_rank` and `ket_rank` **in fixed order**, with no regard for `braket_symmetry()`.

For a braket-symmetric tensor, bra↔ket exchange is a symmetry, so a *half-tensor* `X{a;;x}` (bra-orbital) and its swapped form `X{;a;x}` (ket-orbital) **should** canonicalize identically. They didn't: the ordered `(bra_rank, ket_rank)` gave them different shades → different canonical graphs → different canonical hashes. TNV3 already colors bra/ket slot and bundle vertices symmetrically for such tensors (`is_braket_symm`); the core-vertex shade was the last ordered-by-bra/ket asymmetry.

## Change

- `vertex_painter.cpp`: when `braket_symmetry == Symm`, hash the two ranks as an **unordered** pair (sort them) so the shade is bra↔ket-invariant.
- Regression test (`test_canonicalize.cpp`): a braket-symmetric half-tensor's bra-orbital and ket-orbital forms now share a canonical hash; the non-symmetric forms stay distinct.

## Impact

No effect on tensors with equal bra/ket ranks or non-symmetric braket (the common cases) — the existing canonicalization snapshots are **unchanged** (full unit suite passes, no rebaselines).

## Scope note

This makes the **canonicalizer** recognize braket-symmetric half-tensors as equivalent. It is a prerequisite for, but does not by itself deliver, auto-folding of such half-tensors in the **eval/export** path (e.g. noncovariant-THC export), which additionally requires routing regular (non-ToT) eval leaves through tensor-network canonicalization — a separate, broader follow-up.

### DIFF
--- a/SeQuant/core/tensor_network/vertex_painter.cpp
+++ b/SeQuant/core/tensor_network/vertex_painter.cpp
@@ -26,8 +26,10 @@ std::size_t VertexPainterImpl::to_hash_value(
   // asymmetry, in the core-vertex shade.)
   auto bra_rank = tensor._bra_rank();
   auto ket_rank = tensor._ket_rank();
-  if (tensor._braket_symmetry() == BraKetSymmetry::Symm && bra_rank > ket_rank)
+  if (tensor._braket_symmetry() == BraKetSymmetry::Symm &&
+      bra_rank > ket_rank) {
     std::swap(bra_rank, ket_rank);
+  }
   auto hashes = {hash::value(tensor._label()),
                  hash::value(bra_rank),
                  hash::value(ket_rank),
--- a/SeQuant/core/tensor_network/vertex_painter.cpp
+++ b/SeQuant/core/tensor_network/vertex_painter.cpp
@@ -3,6 +3,8 @@
 #include <SeQuant/core/tensor_network/vertex_painter.hpp>
 #include <SeQuant/core/utility/macros.hpp>
 
+#include <utility>
+
 namespace sequant {
 
 VertexPainterImpl::VertexPainterImpl(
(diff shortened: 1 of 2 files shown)


tests/unit/test_canonicalize.cpp
```diff
@@ -600,3 +601,30 @@ TEST_CASE("canonicalization", "[algorithms]") {
     }
   }
 }
+
+TEST_CASE("braket_symmetric_half_tensor_canonicalization", "[algorithms]") {
+  using namespace sequant;
+  TensorCanonicalizer::register_instance(
+      std::make_shared<DefaultTensorCanonicalizer>());
+  auto isr = sequant::mbpt::make_legacy_spaces();
```
Krzmbrzl: Why do we need legacy index spaces for this?

Krzmbrzl (approved):

tests/unit/test_canonicalize.cpp
evaleev: No reason beyond copy-paste from the existing `canonicalization` test case — this one only uses occupied/unoccupied indices (`i`, `a`), so legacy spaces and PAO are unnecessary. Switched to `make_sr_spaces()`.

SeQuant/core/tensor_network/vertex_painter.cpp
```diff
@@ -14,9 +16,21 @@ VertexPainterImpl::VertexPainterImpl(
...
+  // vertices) must be invariant under swapping the bra and ket ranks. Hash them
+  // as an unordered pair in that case, so e.g. a half-tensor X{a;;x} and its
+  // bra/ket-swapped form X{;a;x} receive the same color and canonicalize to the
+  // same form. (TNV3 already colors bra/ket slot and bundle vertices
+  // symmetrically for such tensors; this removes the last ordered-by-bra/ket
+  // asymmetry, in the core-vertex shade.)
+  auto bra_rank = tensor._bra_rank();
+  auto ket_rank = tensor._ket_rank();
+  if (tensor._braket_symmetry() == BraKetSymmetry::Symm && bra_rank > ket_rank)
+    std::swap(bra_rank, ket_rank);
```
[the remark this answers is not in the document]
evaleev: Braced it to match the rest of the file (f-up commit).

tests/unit/test_canonicalize.cpp
evaleev: Went further: the global test harness already sets the default context to `make_sr_spaces()` (plus cardinal labels), so the test needs no custom context at all now — dropped the setup entirely.

PR #546, opened 2026-06-10
Persistence-aware single-term optimizer cost (n_replay)

evaleev:
## Summary

Adds a persistence-aware cost option to the single-term optimizer. Contractions whose result depends on a "volatile" leaf (e.g. CC amplitudes, recomputed every iteration) can be weighted by an expected replay count `n_replay`, while amplitude-independent ("persistent", computed once) contractions are weighted by 1. This makes the optimizer's cost model match how an iterative solver with intermediate caching actually pays — persistent integral intermediates are built once, volatile work repeats per replay — so factorizations that form amplitude-free integral intermediates up front are correctly preferred.

## API (`OptimizeOptions`)

- `std::function<bool(Tensor const&)> is_volatile_leaf = {}` — marks a leaf tensor as volatile (its value changes between replays). **Empty (default) ⇒ weighting disabled ⇒ behavior identical to before this PR.**
- `unsigned n_replay = 1` — multiplier applied to each volatile contraction's cost; persistent contractions are counted once. Default `1` ⇒ no change. Flops metric only (for Memsize the weight is wrong-signed and is not applied).

## Mechanism

A subset `n` in the bitmask DP is volatile iff it shares a bit with a precomputed `volatile_mask` (built from the volatile leaves in `network.tensors()` order), so the taint propagates through every containing intermediate automatically. The forming-contraction cost is multiplied by `w = (n & volatile_mask) ? n_replay : 1`, yielding a subtree total of `Σ_nodes flops(node) × (n_replay if volatile else 1)`. Applied in both the plain and subnet-CSE cost paths. The CSE path requires the predicate to be invariant under slot/index canonicalization (key on tensor label/structure, not anonymous index identity) — documented on the API.

## Backward compatibility

With the default empty predicate (or `n_replay == 1`), `volatile_mask == 0` and `w == 1` everywhere — existing callers and all `[optimize]` tests are byte-identically unaffected.

## Tests

New `test_optimize.cpp` section on a PPL-shaped motif `g·g·t` with extents chosen so the optimum flips strictly (not tie-dependent): `n_replay=1` buries `t` in an inner intermediate, while large `n_replay` forms the amplitude-free integral first and contracts `t` last. Also asserts the empty-predicate no-op path.

## Downstream

mpqc4's CC solver consumes this for residual factorization (the per-pair 4-PNO particle-particle-ladder integral is built once instead of rebuilt each iteration; ~2× per-iteration speedup observed on PNO-CCSD). The mpqc4 wiring and version repin are a separate change.

### DIFF
--- a/SeQuant/core/optimize/single_term.hpp
+++ b/SeQuant/core/optimize/single_term.hpp
@@ -352,6 +352,15 @@ EvalSequence single_term_opt_impl(TensorNetwork const& network,
 /// \param idxsz An invocable on Index, that maps Index to its dimension.
 /// \param subnet_cse Whether to recognize equivalent subnetworks to try
 /// minimizing the ops counts.
+/// \param is_volatile_leaf Predicate marking a leaf tensor as volatile (its
+/// value changes on every replay); empty disables weighting. The predicate MUST
+/// be invariant under slot/index canonicalization — key on tensor label or
+/// structure, NOT on anonymous index identity — so that two subnetworks deemed
+/// equivalent by the subnet-CSE canonicalization also agree on volatility (the
+/// CSE path stores one cost per canonical subnet). Flops metric only.
+/// \param n_replay Multiplier applied to the cost of each volatile-result
+/// contraction (volatile contractions are re-evaluated on every replay);
+/// persistent contractions are counted once. 1 (default) disables weighting.
 /// \return Optimal evaluation sequence under the chosen cost metric. If there
 ///         are equivalent optimal sequences then the result is the one that
 ///         keeps the order of tensors in the network as original as possible.
--- a/SeQuant/core/optimize/single_term.hpp
+++ b/SeQuant/core/optimize/single_term.hpp
@@ -20,6 +20,7 @@
 
 #include <algorithm>
 #include <bit>
+#include <functional>
 #include <limits>
 #include <type_traits>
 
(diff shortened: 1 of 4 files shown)


tests/unit/test_optimize.cpp
```diff
@@ -199,6 +199,76 @@ TEST_CASE("optimize", "[optimize]") {
...
+                      L" * g_{i2,a2}^{x1}"
+                      L" * t_{a1,a2}^{i1,i2}")
+                      ->as<Product>();
+
+      auto is_t = [](Tensor const& t) { return t.label() == L"t"; };
+
+      OptimizeOptions base;
+      base.idx_to_extent = idxsz;
+
+      // baseline: no predicate => weighting disabled => current behavior
```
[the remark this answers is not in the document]
evaleev: Good catch — fixed in f642d07f3. The comment now says the predicate *is* set here and weighting is neutral because `n_replay == 1`; the empty-predicate no-op is exercised separately via `opts_off` below.

SeQuant/core/optimize/single_term.hpp
```diff
@@ -253,7 +254,8 @@ template <typename CostFn>
   }
 EvalSequence single_term_opt_impl(TensorNetwork const& network,
                                   meta::range_of<Index> auto const& tidxs,
-                                  CostFn&& cost_fn, bool subnet_cse) {
+                                  CostFn&& cost_fn, bool subnet_cse,
+                                  size_t volatile_mask, double n_replay) {
```
Krzmbrzl, line 258: @evaleev why is `n_replay` a `double` here? Isn't it restricted to only integer values?

bimalgaudel: I think this is a misnomer. It should be just a scaling factor to skew the cost, which is the generic case of what `n_replay` is trying to convey. Then the corresponding argument type can remain double. Only @evaleev can tell what's his idea.

PR #548, opened 2026-06-10
fix: handle all-empty-inner ToT arrays

ajay-mk:
An all-empty-inner tensor-of-tensor (a fully screened term in CSV-CC) is logically zero, and `tot_inner_rank()` returns 0 for it. That 0 caused wrong annotation, tripping TA's i`s_tot_index()` check and aborting in `column_symmetrize_ta`, `ResultTensorOfTensorTA::add_inplace`, and `slice_array_over_mode`.

Treat inner-rank-0 as the zero array in each: 
- `symmetrize` returns it unchanged 
- `add_inplace` takes the rank from the populated operand (or no-ops when both are empty) 
- slice builds a zero ToT over the sliced trange
- Adds regression coverage in `test_eval_ta.cpp`

Needed for https://github.com/ValeevGroup/mpqc4/pull/758

### DIFF
--- a/SeQuant/core/eval/backends/tiledarray/result.hpp
+++ b/SeQuant/core/eval/backends/tiledarray/result.hpp
@@ -69,19 +69,22 @@ auto column_symmetrize_ta(TA::DistArray<Args...> const& arr) {
 
   size_t const outer_rank = arr.trange().rank();
 
-  size_t const total_rank = [&] {
-    if constexpr (is_tot) {
-      size_t const inner_rank = tot_inner_rank(arr);
-      if (outer_rank != inner_rank)
-        throw Exception(
-            "ToT symmetrization requires equal outer and inner rank (outer=" +
-            std::to_string(outer_rank) +
-            ", inner=" + std::to_string(inner_rank) + ")");
-      return outer_rank + inner_rank;
-    } else {
-      return outer_rank;
-    }
-  }();
+  if constexpr (is_tot) {
+    size_t const inner_rank = tot_inner_rank(arr);
+    // All-empty-inner ToT (e.g. every CSV pair screened out): no populated tile
+    // to read a rank from, so tot_inner_rank() is 0. It represents zero, and
+    // symmetrizing zero is zero -- return unchanged instead of failing the
+    // equal-rank check below.
+    if (inner_rank == 0) return arr;
+    if (outer_rank != inner_rank)
+      throw Exception(
+          "ToT symmetrization requires equal outer and inner rank (outer=" +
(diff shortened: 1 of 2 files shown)


evaleev (approved):
Reviewed all three sites plus the TA-side assumptions (checked against TA sources). The diagnosis and semantics are right: an all-empty-inner ToT is logically zero, and each site now implements exactly symmetrize(0)=0, t+=0 identity, slice(0)=0.

Verified specifics:
- The hand-built trange in the zero-slice path matches `block()`: tile-aligned, every dim rebased to 0, and the `base` subtraction handles non-zero lobounds. The `s_empty.trange() == s_full.trange()` test on a multi-tile cut at non-zero base is the right regression guard.
- No rank divergence under MPI: `tot_inner_rank()` is collective (`gop.max`) and every new early return branches after the collective, uniformly on all ranks.
- `DistArray{world, trange}` compiles for sparse policies too (routes through `shape_type(1, trange)`; `SparseShape(value_type, TiledRange)` exists), so the "dense even if sparse" comment is accurate.

Suggestions, none blocking:

1. `column_symmetrize_ta`'s early `return arr;` is a shallow handle copy — the "symmetrized" Result aliases the operand's `ArrayImpl`. Safe today since SeQuant mutates result arrays only via expression assignment (which rebinds the handle), but a latent trap if an in-place mutation path is ever added. A one-line comment acknowledging the aliasing (or returning a fresh empty-inner array) would do.
2. On sparse policies the zero-slice path un-screens the term: every tile gets norm 1 and an empty-inner tile is materialized, so downstream contractions iterate tiles sparsity had eliminated. Since `slice_array_over_mode` is on the batched-evaluation hot path, consider an all-zero shape for sparse policies (a legitimately empty sparse array is also TA's zero) and skipping the fill loop. Fine to defer if mixed dense/sparse complicates the batch sum.
3. The `gop.fence()` in the zero-slice path is unnecessary — tiles are set with immediate values by their owners, so readers' futures get fulfilled without a global rendezvous. Minor since it only fires for all-empty arrays.
4. `out.is_local(it.index())` is redundant — `DistArray::begin()/end()` already iterate only local tiles (same in the test helper).
5. Test nit: "populated += empty is a no-op" checks only the surviving inner rank; one data check (values still 1.0) would also guard against a future regression that takes the wrong branch.
6. Optional: in `add_inplace`, `std::max` silently picks a rank if both operands are populated with *different* inner ranks (an upstream logic error). `SEQUANT_ASSERT(r_t == 0 || r_o == 0 || r_t == r_o)` would surface that earlier.

PR #550, opened 2026-06-12
 ITF export: Implement index batching support

Krzmbrzl:


SeQuant/core/index.hpp
```diff
@@ -928,30 +962,31 @@ class Index : public Taggable {
...
+		  return std::nullopt;
+	  }
+
+	  if (*end == '_') {
+		  ++end;
+	  }
+
+	  SEQUANT_ASSERT(end != std::ranges::end(label));
+
+	  std::basic_string_view<std::remove_cvref_t<std::ranges::range_value_t<View>>> view(&(*end), std::ranges::distance(end, std::ranges::end(label)));
```
ajay-mk: Formatting issue? Line 983 is too long. Also surprised the CI didn't flag it.

SeQuant/core/export/itf.hpp
```diff
@@ -289,130 +302,201 @@ class ItfGenerator : public Generator<Context> {
...
   }
 
   void begin_named_section(std::string_view name, const Context &) override {
-    m_generated += "---- code(\"" + std::string(name) + "\")\n";
+    append_line("---- code(\"" + std::string(name) + "\")");
   }
 
   void end_named_section(std::string_view /*name*/, const Context &) override {
-    m_generated += "\n\n";
+    if (!m_currentBatchIndices.empty()) {
```
ajay-mk: This logic for clearing batching indices is repeated in this file again. Maybe factor this out?

SeQuant/core/export/generation_optimizer.hpp
```diff
@@ -514,12 +521,15 @@ class GenerationOptimizer final : public Generator<MainContext> {
     return end;
   }
 
-  void process_operation_queue() {
+  void process_operation_queue(const MainContext &ctx) {
+    IndexSpecificTensorBlockEqualComparator cmp(
```
ajay-mk: `IndexSpecificTensorBlockEqualComparator` is built similarly in other places also in this file. Is the comparator reusable?

SeQuant/core/index.hpp
Krzmbrzl: It's all run through the auto-formatter. So it's as formatted as it gets at least with this specific version of clang-format :D

Krzmbrzl: Interestingly though manual breaking the line survives a pass with clang-format. I'm assuming that this is a bug in the formatter :eyes:

SeQuant/core/export/generation_optimizer.hpp
Krzmbrzl: Yeah, that's a leftover of my initial lazy messing around. Fixed now

PR #555, opened 2026-06-16
feat(mbpt): square-root-normalized `δr`, `δl` projector operators

ajay-mk:
Adds **`δr` and `δl`** square-root-normalized excitation/deexcitation projector and the Normalization machinery they need. Uses `Power` (#513) to represent the square roots. 

**Motivation**: In UCC-EOM the eigenproblem must be Hermitian, which requires the normalization to be split symmetrically between the bra (projector) and ket (operator) sides. To get the standard 1/(c!a!) normalization, it's square root is applied on each side (in both projector and operator). The total stays 1/(c!a!) = 1/(k!)² for a particle-conserving rank-k block; = 1/k! per side. 

#### Normalization conventions
```
Default     →  1/(c! a!)          (standard operator builders)
Implicit    →  no prefactor       (A, S, definition carries normalization)
SquareRoot  →  √(1/(c! a!))       (δr, δl)
```

#### Changelist
- `Normalization` enum is moved to `mbpt` scope and a new member `SquareRoot` is added.
- `OpMaker` supports `SquareRoot` normalization
- `get_unique_indices` now handles `Power` correctly
- Adds an optional normalization argument to `r`, `l`, `R`, `L` (defaulting to `Default`) and `P`, `A`, `S` (defaulting to their intrinsic `Implicit`)
- Defines `δr`/`δl` as `SquareRoot`-normalized `P`: no need to make new definitions, they have the same structure. 
- `eom_r`/`eom_l` now build `R`/`L` with `SquareRoot` normalization and project with `δl`/`δr` instead of bare `P`
- No behavior change for existing `r`/`l`/`R`/`L`/`P`/`A`/`S` callers, the new `Normalization` argument defaults preserves existing behavior.


SeQuant/core/utility/indices.hpp
```diff
@@ -103,6 +103,11 @@ IndexGroups<Container> get_unique_indices(const Variable&) {
   return {};
 }
 
+template <typename Container = std::vector<Index>>
+IndexGroups<Container> get_unique_indices(const Power&) {
+  return {};
```
Krzmbrzl, line 109: To future-proof we might want to add an assertion that power base is not a tensor (as that seems like a generalization that might appear in the future)

SeQuant/domain/mbpt/op.hpp
```diff
@@ -567,6 +567,16 @@ mbpt::qns_t adjoint(mbpt::qns_t qns);
 
 namespace mbpt {
 
+/// @brief Normalization convention used in MBPT Operators
+/// Op of e.g. rank {c,a} by default includes normalization factor of 1/(c! a!).
```
Krzmbrzl: Maybe elaborate a bit more what `c` and `a` are supposed to represent.

SeQuant/domain/mbpt/op.hpp
```diff
@@ -567,6 +567,16 @@ mbpt::qns_t adjoint(mbpt::qns_t qns);
...
+/// For some cases, we want to change that.
+enum class Normalization {
+  Default,    /// Include 1/(c! a!) prefactor
+  Implicit,   /// No prefactor, used for Â and Ŝ since their definition
+              /// includes the normalization
+  SquareRoot  /// Include sqrt(1/c! a!) prefactor
```
Krzmbrzl, line 578: In the literature I have often read the term `Symmetric` for this case. Not sure if this should be preferred here though as I guess `Normalization` itself doesn't imply the presence of a LHS and RHS state/component. 🤔

SeQuant/domain/mbpt/op.hpp
ajay-mk: Done in [2edde54](https://github.com/ValeevGroup/SeQuant/pull/555/commits/2edde54e7d88d841703df252b0c8ca36c95acad7)

SeQuant/core/utility/indices.hpp
ajay-mk: Added `SEQUANT_ASSERT(power.base()->is<Constant>() || power.base()->is<Variable>());`

evaleev (approved):

PR #559, opened 2026-06-19
Generic CostModel for single-term optimization (peak-memory & batched objectives, roofline tie-break)

### LINKED ISSUE #558: optimize: make the optimization objective a pluggable function object
Follow-up from review of #556.

Currently `sequant::OptimizeOptions::objective_function` selects the single-term/top-level optimization cost model from a fixed `enum class ObjectiveFunction { DenseFLOPs, DenseSize }`. @Krzmbrzl suggested making the objective a passed-in **function object** instead, with named default implementations corresponding to today's enum values. That would let downstream users supply custom cost models without modifying SeQuant source.

Scope (separate from the #556 terminology rename):
- Replace the enum dispatch in `single_term.hpp` / `optimize.cpp` with a callable cost interface.
- Provide named default objective objects for the current `DenseFLOPs` / `DenseSize` models.
- Decide how the volatile/footprint weighting (currently gated on `objective_function == DenseFLOPs`) composes with a custom objective.
- Keep the enum as a convenience or remove it.

evaleev:
**TL;DR.** Generalizes single-term optimization into a pluggable `CostModel` framework, so new objectives can be added by writing one model type. The flagship instances delivered here are the peak-memory objectives (`DensePeakSize`/`DensePeakSizeBatched`) and a roofline-based tie-break, which together make local-correlation (PNO/CSV) residuals memory-bounded and keep reused integrals cached.

## Why this PR

In CC-type methods each residual term is a product of many tensors, and the **single-term optimizer** chooses the order in which to contract them. Today it can minimize FLOP count (`DenseFLOPs`) or total operand storage (`DenseSize`) — both essentially order-insensitive and blind to how much memory is *live at once*.

For local correlation methods (PNO/OSV/CSV CCSD, F12) that blindness is the problem: **peak memory, not FLOPs, is the binding resource.** Density-fitted integral construction can transiently materialize an enormous intermediate that still carries a free projected-AO (PAO) index, and the *contraction order* is exactly what decides whether that intermediate is ever formed. A FLOP- or size-minimizing order will happily build it.

This PR teaches the single-term optimizer about peak memory. It adds peak-memory objectives, a small framework to host them (and user-defined ones), a batching model that bounds the peak by slicing the DF index, one batchability policy shared with the runtime evaluator, and — in the most recent commits — the secondary-cost (tie-break) work that makes the optimizer reliably keep reused integrals resident instead of recomputing them.

Everything new is **opt-in and default-neutral**: the existing `DenseFLOPs`/`DenseSize` results are bit-for-bit unchanged, and the roofline tie-break is off by default.

### 1. A `CostModel` framework (refactor)
The subset-lattice + bipartition dynamic program that drives single-term optimization is factored into a single generic driver, `run_single_term_opt<Model>`. Each objective is now a small *model type* that supplies the recurrence and the reconstruction:

- `AdditiveModel` — the existing FLOPs/size objective.
- `PeakModel` — the new peak-memory objective.
- `PeakBatchedModel` — peak memory with batching.

The driver is public and the `CostModel` concept is documented, so a downstream user can define a custom objective and drive it directly. This is also what enables the DRY cleanup below.

### 2. Two peak-memory objectives
- **`DensePeakSize`** — minimizes the *all-co-resident* peak: the maximum, over the evaluation schedule, of the combined size of every simultaneously-live tensor (intermediates **and** resident input leaves). Unlike FLOPs/size, contraction order is a real lever here. Validated against an independent brute-force oracle.
- **`DensePeakSizeBatched`** — models batched evaluation: each index flagged batchable (e.g. the DF/RI auxiliary index) is treated as sliced to a target size, so an intermediate carrying it is materialized one slice at a time. The DP minimizes peak over the worst-case sliced configuration.

### 3. One batchability policy (`BatchPolicy`, `core/batch_policy.hpp`)
A single source of truth — `is_batchable_index`, per-index `batch_target_size`, `is_volatile_leaf` — consumed by **both** the optimizer (so it models what the runtime will do) and the runtime batched evaluator (via a `make_evaluator` adapter over `make_batched_custom_evaluator`). One object, so the two can't drift apart.

### 4. Tie-break quality: Pareto frontier + roofline cost
The peak objective has a primary axis (peak) and a **secondary cost** used to choose among schedules that tie (or nearly tie) on peak. That secondary cost turns out to matter a lot: it decides whether an expensive, *reused* integral is formed once and cached or recomputed every iteration. Two refinements:

- **Pareto-frontier DP** (`peak_flops_tolerance`). A pure peak-min DP with a single per-subset winner cannot reach the global flop-minimum *among* peak-optimal trees (the max-recurrence has no optimal substructure for the secondary objective). The DP now carries a Pareto frontier of non-dominated `(peak, secondary)` points per subset. `peak_flops_tolerance` (default 0.10) then selects the cheapest schedule whose peak is within `(1+tol)` of the minimum — letting the optimizer form, e.g., a persistent 4-index integral that is peak-neutral but much cheaper across the iterations that reuse it.

- **Roofline secondary cost** (`RooflineParams`, default off). The flop-only secondary cost under-weights *bandwidth-bound* contractions (modest FLOPs, large memory traffic), so it treats "fold the amplitude in early and rebuild a large intermediate every iteration" as nearly free. The secondary cost is replaced by a roofline wall-time proxy:

  ```
  cost = max( flops, beta * Q ),   Q = max( |L|+|R|+|Z|, kappa * flops / sqrt(M / c0) )
  ```

  `beta` is the machine balance (FLOPs per element of traffic); `Q` combines compulsory single-pass traffic with the Hong–Kung `sqrt(M)` finite-cache re-read bound; everything is computed on proto-explicit (block-sparse) extents and scaled by the volatile replay weight. Compute-bound (dense) contractions collapse to `flops` — the byte term is inert, so the dense case is unaffected by construction; bandwidth-bound (PNO single-index) contractions are charged their true traffic. `beta` (`machine_balance`) defaults to 0, which reproduces the previous flop-only tie-break exactly. Design write-up: `doc/dev/specs/2026-06-23-roofline-tiebreak-cost.md`.

### 5. DRY cleanup
With every recurrence now living in a model type, the duplicated standalone DP code is removed: `peak_cost`/`peak_cost_batched`/`reconstructed_batched_peak` delegate to the models, and the five now-dead functions (`peak_dp`, `peak_dp_batched`, `single_term_opt_impl`, `single_term_opt_peak_impl`, `single_term_opt_peak_batched_impl`) plus `struct PeakRes` are deleted. Each recurrence exists in exactly one place; the independent `brute_force_min_peak`/`batched_min_peak` oracles stay as cross-checks.

### 6. Backend-aware batch-accumulation footprint penalty (`accumulation_factor`)
A batch-accumulated intermediate (`K += contribution`, summed over aux slices) co-resides the accumulator and the in-flight contribution on an eager backend like TiledArray — a footprint the pure-slice model omits. `accumulation_factor` (on `BatchPolicy` / `OptimizeOptions` / `CostParams`) charges `accumulation_factor * sz(result)` on each node that slices a batchable index, on the primary peak axis. Default `0.0` (backend-agnostic, no factorization change); MPQC defaults it to `1.0` for TA. It asserts at most one batchable index when nonzero — the per-node, once-per-node charge is single-axis-only, which the CC residual (linear in the Hamiltonian => one DF aux per term) always satisfies.

Also removed the throwaway `SEQUANT_BATCH_DECLINE_DEBUG` eval diagnostic added during the investigation.

## Suggested review order
1. `core/optimize/single_term.hpp` — the generic `run_single_term_opt<Model>` driver and the `CostModel` concept.
2. `core/optimize/cost_model.hpp` — `AdditiveModel`, `PeakModel`, `PeakBatchedModel`, and `roofline_op_cost`.
3. `core/optimize/options.hpp` — `OptimizeOptions`, `RooflineParams`, `peak_flops_tolerance`.
4. `core/batch_policy.hpp` + the `make_evaluator` adapter — the shared policy.
5. Tests in `tests/unit/test_optimize.cpp` — oracle cross-checks and the `[osv]` cases that motivate the tie-break.

## Testing
- Full SeQuant unit suite green (6375 assertions); new objectives cross-checked against independent brute-force oracles plus reconstruction-simulation checks; `CostModel` concept conformance (positive and negative).
- New `[bubble]` test: the exchange quadratic bubble `g.t2.t2` at water-20 extents, measuring early-K (held-whole 4-occ/2-PNO integral) vs late-K `t.(gC)` peaks per aux batch size. They share the peak floor (premium <0.1%), so early-K is correct — equal peak plus a persistent build-once flop win, and the memory floor is set by the aux batch size, not the factorization. Confirms `accumulation_factor` and the batch-size knob behave as modeled.
- **Behavior-preserving by default:** `DenseFLOPs`/`DenseSize` unchanged bit-for-bit; with `machine_balance = 0` the peak tie-break is identical to before.
- Also fixes a pre-existing flaky eval test (the two batched-eval paths agree only to `Loose` tolerance, since batched summation order is thread-non-deterministic).
- Downstream: MPQC CSV-CCk (`he10`, batched) reproduces the reference energy to 7e-15.

## End-to-end validation of the tie-break work (PNO-CCSD, water clusters)
On real cluster runs, the roofline secondary cost makes a *low, physically-correct* `volatile_weight` (≈ the actual replay/iteration count) select the cached schedule that the old flop-only tie-break only reached with an artificially inflated `volatile_weight`. On the 20-mer the persistent PPL chain `g·C → (a μ̃|Κ) → (a b|Κ) → W = (ab|cd)` is built once and cached (per-iteration rebuilds of the 36 GB half-transform drop from ~60 to 0), giving roughly **10–17× faster iterations at the same `volatile_weight`**, with bounded memory (aggregate logical working set ~112 GiB; measured node physical ~320 GiB on a 768 GB node).

## Design docs
Specs and plans under `doc/dev/{specs,plans}/` (2026-06).

## Follow-up
MPQC repins `MPQC_TRACKED_SEQUANT_TAG` to this once merged (engine keywords `sequant:optimize:machine_balance` / `:fast_mem_elems` are already staged on the mpqc side).


SeQuant/core/optimize/optimize.cpp
```diff
@@ -36,12 +36,29 @@ ExprPtr opt_pure_product(Product const& prod, OptimizeOptions const& opts) {
...
-        prod, opts.idx_to_extent, subnet_cse, opts.is_volatile_leaf,
-        opts.volatile_weight, opts.footprint_weight);
-  SEQUANT_ASSERT(opts.objective_function == ObjectiveFunction::DenseSize);
-  return opt::single_term_opt<ObjectiveFunction::DenseSize>(
-      prod, opts.idx_to_extent, subnet_cse, opts.is_volatile_leaf,
-      opts.volatile_weight, opts.footprint_weight);
+        prod, opts.idx_to_extent, subnet_cse,
+        opts.batch_policy.is_volatile_leaf, opts.volatile_weight,
+        opts.footprint_weight, opts.batch_policy.is_batchable_index,
+        opts.batch_policy.batch_target_size);
```
Krzmbrzl: I guess it would be easier to just pass `opts` in its entirety into the function instead of every member separately, no?

doc/dev/plans/2026-06-18-cost-model-batch-aware-phase2-plan.md
Krzmbrzl, line 1: These files should be removed before merging.

evaleev: these files contain valuable details about the design.

Krzmbrzl: I feel like they would have to be rewritten significantly in order to be actually useful documentation.

evaleev: probably, but no bandwidth available to write for human consumption. This is still better than nothing. This type of knowledge is not extractable from the code afterwards.

SeQuant/core/optimize/optimize.cpp
evaleev: Done in f9b01dd72: the five cost-model knobs (`is_volatile_leaf`, `volatile_weight`, `footprint_weight`, `peak_flops_tolerance`, `roofline`) are now bundled into a `CostParams` struct, and `opt_pure_product` builds one `CostParams` from `opts` and forwards it to all four `single_term_opt` arms. I kept `CostParams` distinct from `OptimizeOptions` rather than passing `opts` wholesale so that the templated `opt::single_term_opt` entry points stay decoupled from the public options struct (they have no need for `reorder`, `CSE`, `objective_function`, etc., and shouldn't pull in that dependency). Behavior is bit-identical (full unit suite: 6375 assertions, 56 cases).

SeQuant/core/optimize/single_term.hpp
```diff
@@ -230,161 +409,64 @@ inline SubnetMetadata build_subnet_metadata(
...
-    results[n].sequence =
-        (lseq[0] < rseq[0] ? concat(lseq, rseq) : concat(rseq, lseq)) |
-        ranges::to<EvalSequence>;
-    results[n].sequence.push_back(-1);
-  }
+// The additive arms of single_term_opt<Metric> below are routed through the
+// generic CostModel driver. cost_model.hpp includes this header (for the DP
+// helpers + EvalSequence + OptRes); the include guards make the cycle a no-op,
+// and every helper cost_model.hpp needs is defined above this point.
+#include <SeQuant/core/optimize/cost_model.hpp>
```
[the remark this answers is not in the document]
evaleev: Fixed in 3e5ee5c2a by splitting the shared DP helpers (cost counters, `OptRes`/`EvalSequence`, `init_results`, the subset/bipartition scaffolding) into a new `single_term_detail.hpp`. The dependency graph is now a DAG: `single_term_detail.hpp` (no optimize deps) <- `cost_model.hpp` <- `single_term.hpp`. `cost_model.hpp` includes only the detail header, and `single_term.hpp` includes `cost_model.hpp` before defining the public `single_term_opt<Metric>` entry points. `cost_model.hpp` is now standalone-includable — verified by compiling a TU that does `#include <SeQuant/core/optimize/cost_model.hpp>` first with the real build flags. Also registered `cost_model.hpp` (previously unlisted) and the new header in CMakeLists.

SeQuant/core/optimize/single_term.hpp
```diff
@@ -31,6 +35,47 @@ concept has_index_extent = std::is_invocable_r_v<size_t, F, Index const&>;
...
+      std::size_t k = 0;
+      for (auto const& o : tot_idxs.inner)
+        if (o.proto_indices() == c.proto_indices()) ++k;
+      mem *= inner_pow(c, k);
+    }
+  } else {
+    mem = ranges::accumulate(tot_idxs.inner, mem, std::multiplies{}, ixex);
+  }
+  return mem;
+}
```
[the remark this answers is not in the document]
evaleev: Fixed in 3e5ee5c2a. Added an `option_engaged()` helper: for `std::function`/function pointers it reports the engaged (non-empty) state via contextual conversion to bool; for any other callable (e.g. a plain lambda, which has no empty state) it returns true. `inner_aware_volume` now does `if (option_engaged(inner_pow))`, so a caller may pass a bare lambda for `inner_pow` as well as an optionally-empty `std::function`.

SeQuant/core/eval/eval.hpp
```diff
@@ -1259,6 +1267,36 @@ template <typename F, typename IndexPredicate = accept_any_index,
...
+                                  ScopeGuardFactory make_scope_guard = {}) {
+  auto is_volatile_node = [p = policy.is_volatile_leaf](auto const& n) -> bool {
+    if (!n.leaf() || !n->is_tensor()) return false;
+    return p && p(n->as_tensor());
+  };
+  return make_batched_custom_evaluator(
+      std::move(yielder), policy.batch_target_size, policy.is_batchable_index,
+      std::move(make_scope_guard), std::move(is_volatile_node),
+      policy.persistent_only);
+}
```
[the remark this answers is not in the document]
evaleev: Fixed in 3e5ee5c2a. `make_evaluator` now checks both `policy.is_batchable_index` and `policy.batch_target_size`; if either is empty it substitutes predicates that decline batching (an accept-nothing predicate, so `batch_axis` returns `nullopt` and `target_batch_size` is never called) rather than forwarding the empty `std::function`s. This matches the BatchPolicy docs (empty => no batching) and avoids the `std::bad_function_call` at evaluation time.

doc/dev/plans/2026-06-18-cost-model-batch-aware-phase2-plan.md
Krzmbrzl: If it's not intended for humans, I vote for moving it outside of `docs` (in something like `agent-docs`) or put it in a subdirectory like`docs/for_agents`. Also, the files could be renamed to something a bit more meaningful

PR #567, opened 2026-06-29
eval: add out-of-band SlotSymmetry descriptor carrier on EvalExpr (Phase 0)

evaleev:
## WARNING
This is an experimental sandbox playing in which should lead to sufficient understanding of the problem. The max goal is to figure out the algorithm for symmetry detection and min goal is to generate enough unit tests. The initial agentic-generated solution is incredibly naive, but hopefully subsequent iterations will produce something more robust.

## Summary

Phase 0 of per-intermediate permutational-symmetry deduction: add the **data carrier only**, no deduction logic yet.

- New `SeQuant/core/eval/slot_symmetry.hpp`: a `SlotSymmetry` descriptor struct
  - `column_groups` (over matched (bra[c], ket[c]) columns), `bra_groups`, `ket_groups`; each group carries a sign
  - `empty()` and an order-insensitive `operator==`
- `EvalExpr` gains a default-constructed, **out-of-band** `slot_symmetry_` member plus a `slot_symmetry()` accessor

### Out-of-band constraint (encoding 1a)
The new field is write-default-only and is intentionally **not** referenced by any identity/serialization path: `canonicalize_slots`, `EvalExpr` hashing, the `TensorNetwork` bliss-graph builder, or export. No intermediate result tensor's symmetry tag is touched (`make_tensor_wo_symmetries` is unchanged). `grep` confirms `slot_symmetry_`/`slot_symmetry()` appear only in the three intended files. Later phases will populate and consume it.

## Test plan
TDD: failing test first (RED: `slot_symmetry.hpp` not found), then implement (GREEN).

- New `tests/unit/test_slot_symmetry.cpp` (tag `[slot_symmetry]`): default-empty descriptor, `operator==`, carrier present + empty on a leaf `EvalExpr`, and non-empty != empty. **5 assertions, pass.**
- Regression guard (Debug): `[optimize],[EvalExpr],[EvalNode]` -> **486 assertions in 11 test cases, all pass**; `[slot_symmetry]` passes. (This branch also includes the `[bubble]` Debug-timeout fix as its parent commit, so the optimize suite now completes in Debug.)
- Also rebuilt + ran clean in Release (`[optimize]` 323 assertions; `[slot_symmetry],[EvalExpr],[EvalNode]` 170 assertions).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -440,6 +440,8 @@ set(SeQuant_eval_src
         SeQuant/core/eval/eval_node_compare.hpp
         SeQuant/core/eval/result.cpp
         SeQuant/core/eval/result.hpp
+        SeQuant/core/eval/slot_symmetry.cpp
+        SeQuant/core/eval/slot_symmetry.hpp
         SeQuant/core/eval/fwd.hpp
 )
 
--- a/SeQuant/core/eval/eval_expr.cpp
+++ b/SeQuant/core/eval/eval_expr.cpp
@@ -475,8 +475,16 @@ EvalExprNode binarize(Sum const& sum, IndexSet const& uncontract,
           1,                                                //
           h,                                                //
           nullptr};
-      EvalOpSetter{}.set_slot_symmetry(
-          res, intersect(left.slot_symmetry(), right.slot_symmetry()));
+      // intersect compares groups by integer slot position; this is valid
+      // because all summands share the same external-slot layout (the Sum
+      // result is built from left.canon_indices()). Only propagate the
+      // descriptor when the result preserves the operand's bra/ket layout
+      // (not the case under merge_indices mode, I1 fix; I3 invariant).
+      if (res.as_tensor().bra_rank() == left.as_tensor().bra_rank() &&
+          res.as_tensor().ket_rank() == left.as_tensor().ket_rank())
+        EvalOpSetter{}.set_slot_symmetry(
+            res, intersect(left.slot_symmetry(), right.slot_symmetry()));
+      // else: leave slot_symmetry empty (default-constructed).
       return res;
     } else {
       return {EvalOp::Sum,              //
@@ -547,10 +555,14 @@ EvalExprNode binarize(Product const& prod, IndexSet const& uncontract,
           1,                                                //
           h,
           nullptr};
-      // A scalar factor cannot break permutational symmetry: the tensor
-      // operand's descriptor passes through unchanged (its slot layout is
-      // preserved by make_tensor_wo_symmetries).
-      EvalOpSetter{}.set_slot_symmetry(res, tl->slot_symmetry());
+      // A scalar factor cannot break permutational symmetry: pass through the
+      // tensor operand's descriptor only when the result preserves the
+      // operand's bra/ket layout (not the case under merge_indices mode, where
+      // all indices collapse into aux and the slot positions are meaningless).
+      if (res.as_tensor().bra_rank() == t.bra_rank() &&
+          res.as_tensor().ket_rank() == t.ket_rank())
+        EvalOpSetter{}.set_slot_symmetry(res, tl->slot_symmetry());
+      // else: leave slot_symmetry empty (default-constructed).
       return res;
     } else {
       // tensor * tensor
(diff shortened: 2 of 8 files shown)


Krzmbrzl:
Wouldn't it make more sense to make `Tensor` objects be able to store this partial symmetries (in addition to the current ones)? 'cause ultimately that's what we'll want and need for generalized symmetry support. If we introduce it as a separate property of an eval node, I can kinda see this becoming an issue once tensors are upgraded to general symmetries :thinking: 

In fact, the most reusable way to implement this is to make use of `libPerm` and just store a proper permutation group from which we can query the subset of permutation actions that we currently care about (or rather: know what to do with). This would probably maximize compatibility w.r.t. a future full symmetry generalization.

evaleev:
> Wouldn't it make more sense to make `Tensor` objects be able to store this partial symmetries (in addition to the current ones)? 'cause ultimately that's what we'll want and need for generalized symmetry support. If we introduce it as a separate property of an eval node, I can kinda see this becoming an issue once tensors are upgraded to general symmetries 🤔

When we are ready to support general symmetries it will make sense to reuse this in AbstactTensor API and Tensor.

> In fact, the most reusable way to implement this is to make use of `libPerm` and just store a proper permutation group from which we can query the subset of permutation actions that we currently care about (or rather: know what to do with). This would probably maximize compatibility w.r.t. a future full symmetry generalization.

Presumably you mean symmetry is specified by a representation. I'm not sure over what actually (since numbers alone are not enough, presumably over operators that include complex conjugation etc.)

I do not think I'm able to bite all of this off at once. Consider this a modeling exercise that will inform future design decisions but meanwhile will be sufficient for some improvements. Since it took us 3 attempts to get TN canonicalizer, I assume this will take at least this many iterations.

Krzmbrzl:
> When we are ready to support general symmetries it will make sense to reuse this in AbstactTensor API and Tensor.

But why not put it there from the get-go? It seems that would implicitly also cover the eval expr case by means of every node having a result tensor.

> Presumably you mean symmetry is specified by a representation. I'm not sure over what actually (since numbers alone are not enough, presumably over operators that include complex conjugation etc.)

No, I mean symmetry represented as (basically) a list of individual permutations (including sign). Not that much different from storing indices of permutable elements, really.

From a superficial glance, it seems like the new symmetry object introduced by this PR mostly re-invents what libPerm already does. Perhaps it would we worth it to have a quick in-person chat about this?

evaleev:
> But why not put it there from the get-go? It seems that would implicitly also cover the eval expr case by means of every node having a result tensor.

I do not have a complete picture in my mind of how these concepts interplay. E.g. what are the "column" arrangement of the intermediate's slots? Do we need to deduce the column structure of tensor network? etc. 

> No, I mean symmetry represented as (basically) a list of individual permutations (including sign). Not that much different from storing indices of permutable elements, really.

General symmetry handling will need to consider more general model than this (which is OK for real tensors with simple permutational symmetries), e.g. need for permutations to result in complex conjugation + nonabelian symmetries (in this case representation of the symmetry is not representable by plus or minus 1).

Krzmbrzl:
Extension to complex conjugation is only an extension of the permutation class in libPerm. Also, libPerm doesn't make assumptions about Abelian-ness so non-abelian should not cause problems.

> I do not have a complete picture in my mind of how these concepts interplay. E.g. what are the "column" arrangement of the intermediate's slots? Do we need to deduce the column structure of tensor network? etc.

With libPerm we could represent individual index permutations so the tensor layout doesn't have to be anything specific in order to represent what we might want to refer to as a "column symmetry" in any given context.

evaleev:
> With libPerm we could represent individual index permutations so the tensor layout doesn't have to be anything specific in order to represent what we might want to refer to as a "column symmetry" in any given context.

it still seems that `AbstractTensor`/`Tensor` is a wrong abstraction here, as it only represents top-level (non-proto) slots in the PNO/CSV case. Here we presumably want to represent all slots (including those on CSV indices).

P.S. another potential concern raised by an agentic review: to deduce symmetry of a Sum we need to determine the intersection of group representations of multiple terms. It says this is nontrivial for a general permutation case assumed by libPerm.

PR #568, opened 2026-07-01
eval/optimize/dryrun: nested aux+PAO batching, perf-first objective, dry-run cost-profile backend, rank-general CSV moments

evaleev:
Eval/optimize/dry-run infrastructure for costing, sizing, and batching CSV-CCk / PNO-CCSD factorizations, motivated by the C60 PNO-CCSD OOM (the dense free-μ̃ `(g·C)(g·C)` giant). This branch grew well past its original "observe-only predicted-footprint hook" scope (that hook has since been removed — see below). Paired MPQC PR: ValeevGroup/mpqc4#783.

### 1. Nested multi-axis (aux + PAO + occupied) batching in the evaluator
- `EvalExpr.batch_axes_`: `binarize` stamps per-node sliced-sets; the batched evaluator prefers the per-node annotated batch axis.
- Recursive re-entry on the batched scratch realizes nested multi-axis batching (one batch loop per annotated axis); screening relaxation scales by the product of batch counts.
- **External-occupied (spectator) forest batching**: a free/spectator external occupied axis -- carried on the root but contracted at no node (e.g. a PNO/CSV composite's occupied protoindex) -- is sliced by an OUTER forest loop wrapping the whole per-term evaluation, with the aux+PAO intra-term batching nested inside. `optimize()` recognizes such occupied protoindices as batchable and emits the choice as a forest-level `ExternalBatchAxis` signal (domain-neutral `BatchPolicy` config). This lets PNO-CCSD batch over **occupieds**, not just aux/PAO, to bound the dense free-mu-tilde `(g.C)(g.C)` giant.
- `OptimizeOptions.term_batch_axes` surfaces per-node sliced-sets per summand in RPN order.

### 2. Threshold-gated batched selection in `optimize()`
- `PeakBatchedModel` / `reconstructed_batched_peak` oracle prices the batch contribution buffer; `peak_threshold` (bytes) plumbed through `BatchPolicy` / `CostParams`.
- Threshold-gated selection: min FLOPs s.t. peak ≤ P_max, min-peak fallback. `accumulation_factor` allowed for m≥2 (nested), gated by the DP==oracle identity. Sliced-axis model target floored to ≥1 to match the runtime.

### 3. Perf-first `DenseTimeSpace{,Batched}` single-term objective
- New objective selecting by FLOPs-then-peak (perf-first, peak-second), so a peak threshold no longer forces a FLOPS-catastrophic-but-sliceable factorization (the C60 4-PAO AO-integral pathology). Design spec + implementation plan included; C60 perf-first vs peak-first comparison test.

### 4. Dry-run cost-profile eval backend
- A zero-data `Result` backend + `CostModel` that replays a factorized IR and reports memsize + FLOPs + exec cost (a pluggable sizing oracle; dense by default), not just size.
- `CostProfile` struct + reusable `cost_profile()` entry point; `[dryrun-perf]` / `[dryrun-trace]` routed through it. Opt-in scratch-fold peak sink for a faithful batched-replay peak; gated cache with batchable-axis veto and per-term reset.
- `SizeRegime`: per-space extents + per-rank CSV moment tables (power means). Post-transform (μ̃ + Κ) C60 residual fixtures and DP-inspection mechanics; power-mean sizing contract locked with a regression test.
- **Rank-general CSV moment dispatch**: `inner_pow` dispatches by cluster rank (1→OSV, 2→PNO, ≥3→a `csv_moment_by_rank` table if present, else the PNO-table fallback), so CSV-CCSDT triples are sized by their own domain. Pairs with the MPQC rank-general `CSVSizeMoments`.

### 5. Index-space registry: occupancy predicates robust to non-physical spaces
Enabling the occupied-spectator scan in `optimize()` against a real CSV/PNO registry surfaced a latent registry bug. `is_pure_occupied` / `is_pure_unoccupied` / `contains_occupied` / `contains_unoccupied` reduced an index's quantum numbers to physical-particle (spin) attributes and then looked up the vacuum-(un)occupied space via a *throwing* retrieve, so an index whose reduced qns had no registered occupied space threw instead of answering the occupancy question. Two coordinated fixes:
- **Physical spaces (AO, PAO) are spin-*independent*, not spin-*less*.** They span real particle states, so they must carry the convention's `spin_any` in the physical (spin) sector alongside their LCAO trait bit. `add_ao_spaces` / `add_pao_spaces` previously registered them with only the trait bit (spin bits empty), so `physical_particle_attributes()` reported no spin and the occupancy lookup queried a spin flavor with no occupied space. These now take an explicit `spin_any` parameter (`Spin::null` for `SpinConvention::Legacy`, else `Spin::any`, matching `make_*_spaces`) and register `spin_any | trait`.
- **Non-physical auxiliary spaces (density-fitting, batching) legitimately carry no spin;** asking whether they are occupied must answer *false*, not throw. The four predicates now route through non-throwing helpers (`vacuum_occupied_type_or_null` / `complete_type_or_null` / `vacuum_unoccupied_type_or_null`) that return a null type when no space is registered at the reduced qns. Byte-identical for physical inputs.

Unit-test callers of `add_ao_spaces` / `add_pao_spaces` were updated to pass the `spin_any` matching each test's registry convention (Default/None → `Spin::any`, Legacy → `Spin::null`).

### 6. Fixes / misc
- Fix stack-use-after-scope in `tot_inner_rank`.
- Doc: rename the eval backend brand `CostToken`/`DryRun` for clarity; reframe the spec as cost-model replay.

## Note on history
The original observe-only predicted-footprint (predict) hook on `CacheManager` was **added and then removed** within this branch (commits `a19dbfb5f` → `bc5bd6cb5`); it nets to zero. The dry-run cost-profile backend above is its principled replacement — it predicts the whole factorized IR up front rather than per-op during materialization. This PR's title/body have been updated accordingly.

## Validation
`[dryrun]` / `[dryrun-df]` / `[dryrun-perf]` / `[optimize]` suites pass; new rank-general dispatch regression test passes. The registry fix keeps the full space/mbpt suite green (`[elements]`, `[mbpt]`, `[mbpt/cc]`, `index_space`, `spin`, `[algorithms]` incl. the Legacy-convention PAO canonicalization test) and `[optimize]` (which exercises the PAO+DF `compute_external_batch_axis` emission path); MPQC `he10-csv-cck-2-pao` correlation energy is unchanged vs the pre-fix binary to ~1e-13. Branch base is `master`.


evaleev:
Superseded by #583 (multimode-batched-eval), which consolidates this branch's work plus everything built on top (external-occ scatter, order-aware placement, slice-on-use, lifetime-mask per-level placement) into 6 clean commits rebased on master. Closing in favor of #583.

PR #574, opened 2026-07-07
Lift the >2-tensor limit on external auxiliary hyperindices (TNv3 canonicalization + BTAS batched eval)

evaleev:
## Summary

SeQuant already supports **hyperindices** — indices shared among more than two tensors — as exercised by the existing tensor-hypercontraction / THC tests and the order-3 `T{a1;i1} T{a2;i1} T{a3;i1}` test. This PR does **not** add hyperindex support; it lifts a limitation that only shows up at higher order.

In every existing test the hyperindex is either **contracted** (an internal/anonymous mode — skipped by the named-index handling in canonicalization) or **external but shared by at most two tensor slots**. Neither reaches the one code path that assumed a *named* index's edge connects at most two vertices, so the limit was never hit — the existing coverage tops out at order-3 *contracted* hyperindices.

Laplace-transform MP2 is the first case with an **external hyperindex of order > 2**: a batching/quadrature index `z1` rides in the aux slot of every factor *and* of the result `E{;;z1}` (order 9 here). That tripped the assertion, and — once past it — could not be evaluated. Both layers are addressed:

### 1. Tree building (TNv3 canonicalization)

`TensorNetworkV3::canonicalize_slots` classified each **named** index by its edge's vertex count, handling 1 (an open bra/ket/aux slot) or 2 (an internal contraction that is also named, e.g. a protoindex) and asserting `vertex_count() == 2` otherwise. A named index shared by >2 tensor slots forms a hyperedge with >2 vertices and tripped that assert.

Fix: classify an all-aux, >2-vertex named edge as a `TensorAux` slot. A non-auxiliary high-order hyperindex has no well-defined slot type and is rejected with an explicit `throw` (in every build config, not only assertion-enabled ones).

### 2. Numerical evaluation

Each contraction node over such an index is a **batched (Hadamard) contraction**: the index is present in the left operand, the right operand *and* the result, so it must be iterated, not summed.

- **BTAS** — `btas::contract` has no batch category (an index in both operands is always summed), so it aborted. `ResultTensorBTAS::prod` now detects batch axes (`batch_indices`) and routes through a new `batched_contract()` that permutes the batch axes to the front (making each slice contiguous in row-major storage) and contracts each slice — falling back to a dot or a scaling for the degenerate slice shapes.
- **TiledArray** — no backend change needed; `einsum` supports batch axes natively. Only the layer-1 fix was required.

## Tests

- `binarize_highorder_aux_hyperindex` (`test_eval_btas.cpp`) — reproduces the Laplace-MP2 energy expression; asserts the eval tree builds and carries `z1` in the aux slot of every tensor node. Confirmed red→green against the fix.
- `btas_batched_contract` — unit-tests `prod()` on every slice shape (contract+batch, dot+batch, scalar·tensor+batch, non-canonical result order).
- `eval_btas_batched_over_aux` — end-to-end `binarize + evaluate` of a 3-tensor batched-over-aux contraction against a hand-computed reference.
- `test_eval_ta.cpp` non-covariant section — end-to-end TA evaluation of the same class of expression against a `TA::einsum` reference.

Full BTAS unit suite: 7194 assertions / 65 cases pass. TA `eval_with_tiledarray` case passes with the new test.

## Notes

- `IndexSlotType::TensorAuxAux` is removed — it had no users, and a high-order auxiliary hyperindex is classified as `TensorAux`, so there is no distinct aux-aux slot type.
- The batch-extent agreement of the two operands is asserted in `batched_contract` before slicing.

### DIFF
--- a/SeQuant/core/eval/backends/btas/result.hpp
+++ b/SeQuant/core/eval/backends/btas/result.hpp
@@ -196,8 +196,16 @@ T batched_contract(typename T::numeric_type alpha, T const& L, Annot const& la,
     }
     return std::pair{std::move(e), sz};
   };
+  // both operands were permuted to [batch..., rest...], so their leading nb
+  // axes are the batch axes in the same order; a well-formed annotation gives
+  // them identical extents. Guard against a malformed one before slicing (P is
+  // taken from the left operand below).
   std::size_t P = 1;
-  for (std::size_t i = 0; i != nb; ++i) P *= Lp.extent(i);
+  for (std::size_t i = 0; i != nb; ++i) {
+    SEQUANT_ASSERT(Lp.extent(i) == Rp.extent(i) &&
+                   "batched_contract: operands disagree on a batch extent");
+    P *= Lp.extent(i);
+  }
   auto const [rlext, rlsz] = tail_extents(Lp, nb);
   auto const [rrext, rrsz] = tail_extents(Rp, nb);
 
--- a/SeQuant/core/eval/backends/btas/result.hpp
+++ b/SeQuant/core/eval/backends/btas/result.hpp
@@ -128,6 +128,153 @@ inline void log_btas(Args const&... args) noexcept {
   log_result("[BTAS] ", args...);
 }
 
+/// \brief The batch (Hadamard) axes of a binary contraction: labels that occur
+///        in the left operand, the right operand AND the result.
+///
+/// btas::contract classifies every index as free-left, free-right or contracted
+/// (an index shared by both operands is *always* summed); it cannot represent a
+/// batch axis -- an index carried through the contraction rather than summed,
+/// e.g. an auxiliary/quadrature index in Laplace-transform MP2 or tensor
+/// hypercontraction. Such axes are detected here and handled by
+/// batched_contract() below.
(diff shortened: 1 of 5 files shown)


evaleev:
Thanks for the review. Addressed in 8474a9d28:

1. **`v3.cpp` — non-aux hyperindex silently classified under `IGNORE`** (valid): changed the `SEQUANT_ASSERT` to an unconditional `throw Exception`, so a non-auxiliary high-order hyperindex hard-errors in every build config rather than falling through to `TensorAux` when assertions are compiled out. `all_aux` is now read unconditionally (the `[[maybe_unused]]` is gone).

2. **`result.hpp` — `P` derived only from the left operand** (valid): added a `SEQUANT_ASSERT` that the two operands agree on every batch extent (their leading `nb` axes after the batch-first permute) before computing `P` and slicing.

3. **`slot.hpp` — keep `TensorAuxAux` as a legacy alias** (declined): the enumerator had no users anywhere in the tree, and its removal was explicitly requested. A high-order auxiliary hyperindex is classified as `TensorAux`, so there is no distinct aux-aux slot type to preserve. Leaving it removed.

PR #578, opened 2026-07-19
feat: CC RDMs

### LINKED ISSUE #572: Implement RDMs in `mbpt::CC` class
Add support for reduced density matrices (RDMs) to the `mbpt::CC` class.

All of the required infrastructure is already in place, and the RDMs can already be derived. This just needs to be integrated into the class and made to work for all supported ansätze.

ajay-mk:
Resolves #572. Needed for for the real-time project. 

The equations are numerically verified in MPQC (T and U ansatz, SO/CS, CCSD and CCSDT). MPQC PR: https://github.com/ValeevGroup/mpqc4/pull/789

### DIFF
--- a/SeQuant/domain/mbpt/models/cc.cpp
+++ b/SeQuant/domain/mbpt/models/cc.cpp
@@ -222,11 +222,13 @@ ExprPtr CC::rdm(size_t rank, std::optional<size_t> comm_rank) const {
   // quasi-creators must be absorbed by ã's 2r plus Λ's 2N legs (k <= r + N).
   // Unitary ansatz: T⁺ contracts with T, the expansion never terminates, so
   // there is no safe default; use the engine's hbar_comm_rank (the ctor
-  // guarantees it is set).
+  // guarantees it is set). The traditional branch takes lst_options()'s
+  // connected-product form; the {ã,t} connectivity handed to ref_av below is
+  // what makes it equivalent to the explicit commutator.
   const auto commutator_rank = comm_rank.value_or(
       unitary() ? hbar_comm_rank_.value() : std::min(2 * rank, rank + N));
-  auto bar = mbpt::lst(replacer, T(N, skip_singles()), commutator_rank,
-                       {.unitary = unitary()});
+  auto bar =
+      mbpt::lst(replacer, T(N, skip_singles()), commutator_rank, lst_options());
 
   // 3. reference expectation value with the ansatz's left wavefunction:
   // γ = <0|(1+Λ) e^{-σ} ã e^{σ}|0>; the unitary ansatz drops Λ.
--- a/SeQuant/domain/mbpt/models/cc.cpp
+++ b/SeQuant/domain/mbpt/models/cc.cpp
@@ -220,6 +220,10 @@ ExprPtr CC::rdm(size_t rank, std::optional<size_t> comm_rank) const {
   // largest number of T's that can survive <0|(1+Λ) N̄|0>:
   // Unitary ansatz: T⁺ contracts with T, the expansion never terminates, so
   // there is no safe default; use the engine's hbar_comm_rank.
+  if (unitary())
+    SEQUANT_ASSERT(hbar_comm_rank_ &&
+                   "hbar_comm_rank must be specified for unitary ansatz in "
+                   "CC::rdm");
   const auto commutator_rank = comm_rank.value_or(
       unitary() ? hbar_comm_rank_.value() : std::min(2 * rank, rank + N));
   auto Nbar = mbpt::lst(N_op, T(N, skip_singles()), commutator_rank,
--- a/SeQuant/domain/mbpt/models/cc.cpp
+++ b/SeQuant/domain/mbpt/models/cc.cpp
@@ -211,13 +211,13 @@ std::vector<ExprPtr> CC::λ() const {
 }
 
 ExprPtr CC::rdm(size_t rank, std::optional<size_t> comm_rank) const {
-  // 1. rank-`rank` number operator N = {a†_{p1..pr} a_{q1..qr}};(see op::N).
-  // Qualified with op:: because `N` (the cluster rank) is a member.
-  auto N_op = op::N(rank);
+  // 1. replacement operator {ã^{p_1..p_r}_{p_{r+1}..p_{2r}}} (see op::ã); its
+  // indices are free, so they become the free indices of γ.
+  auto replacer = op::ã(rank);
 
-  // 2. similarity transform N̄ = e^{-σ} N e^{σ} (σ = T, or T−T⁺ for unitary).
-  // Traditional ansatz: the expansion terminates exactly, so the default is the
-  // largest number of T's that can survive <0|(1+Λ) N̄|0>:
+  // 2. similarity transform e^{-σ} ã e^{σ} (σ = T for traditional and σ = T−T⁺
+  // for unitary). Traditional ansatz: the expansion terminates exactly, so the
+  // default is the largest number of T's that can survive the VEV below:
   // Unitary ansatz: T⁺ contracts with T, the expansion never terminates, so
   // there is no safe default; use the engine's hbar_comm_rank.
   if (unitary())
@@ -226,15 +226,15 @@ ExprPtr CC::rdm(size_t rank, std::optional<size_t> comm_rank) const {
                    "CC::rdm");
   const auto commutator_rank = comm_rank.value_or(
       unitary() ? hbar_comm_rank_.value() : std::min(2 * rank, rank + N));
-  auto Nbar = mbpt::lst(N_op, T(N, skip_singles()), commutator_rank,
-                        {.unitary = unitary()});
+  auto bar = mbpt::lst(replacer, T(N, skip_singles()), commutator_rank,
+                       {.unitary = unitary()});
 
   // 3. reference expectation value with the ansatz's left wavefunction:
-  // γ = <0|(1+Λ) N̄|0>  (traditional);  <0|N̄|0>  (unitary, no Λ).
-  auto expr = unitary() ? Nbar : simplify((1 + Λ(N, skip_singles())) * Nbar);
+  // γ = <0|(1+Λ) e^{-σ} ã e^{σ}|0>; the unitary ansatz drops Λ.
+  auto expr = unitary() ? bar : simplify((1 + Λ(N, skip_singles())) * bar);
   const auto connect =
       unitary() ? mbpt::OpConnections<std::wstring>{}
-                : mbpt::OpConnections<std::wstring>{{L"N", L"t"}, {L"t", L"N"}};
+                : mbpt::OpConnections<std::wstring>{{L"ã", L"t"}, {L"t", L"ã"}};
   auto gamma = this->ref_av(expr, connect);
 
   return gamma;
--- a/SeQuant/domain/mbpt/models/cc.cpp
+++ b/SeQuant/domain/mbpt/models/cc.cpp
@@ -44,12 +44,12 @@ CC::CC(size_t n, const Options& opts)
       hbar_comm_rank_(opts.hbar_comm_rank),
       pertbar_comm_rank_(opts.pertbar_comm_rank) {
   if (unitary())
-    SEQUANT_ASSERT(hbar_comm_rank_ &&
+    SEQUANT_ASSERT(hbar_comm_rank_,
                    "CC: hbar_comm_rank is required for unitary ansatz");
   if (ansatz_ == Ansatz::oT || ansatz_ == Ansatz::oU)
-    SEQUANT_ASSERT(
-        skip_singles_ &&
-        "CC: skip_singles must be true for orbital-optimized ansatz");
+    SEQUANT_ASSERT(skip_singles_,
+                   "CC: skip_singles must be true for orbital-optimized "
+                   "ansatz");
 }
 
 CC::Ansatz CC::ansatz() const { return ansatz_; }
@@ -217,13 +217,12 @@ ExprPtr CC::rdm(size_t rank, std::optional<size_t> comm_rank) const {
 
   // 2. similarity transform e^{-σ} ã e^{σ} (σ = T for traditional and σ = T−T⁺
   // for unitary). Traditional ansatz: the expansion terminates exactly, so the
-  // default is the largest number of T's that can survive the VEV below:
+  // default is the largest number of T's that can survive the VEV below: each T
+  // must consume at least one of ã's 2r legs (k <= 2r), and T's 2k
+  // quasi-creators must be absorbed by ã's 2r plus Λ's 2N legs (k <= r + N).
   // Unitary ansatz: T⁺ contracts with T, the expansion never terminates, so
-  // there is no safe default; use the engine's hbar_comm_rank.
-  if (unitary())
-    SEQUANT_ASSERT(hbar_comm_rank_ &&
-                   "hbar_comm_rank must be specified for unitary ansatz in "
-                   "CC::rdm");
+  // there is no safe default; use the engine's hbar_comm_rank (the ctor
+  // guarantees it is set).
   const auto commutator_rank = comm_rank.value_or(
       unitary() ? hbar_comm_rank_.value() : std::min(2 * rank, rank + N));
   auto bar = mbpt::lst(replacer, T(N, skip_singles()), commutator_rank,
@@ -248,13 +247,9 @@ std::vector<ExprPtr> CC::tʼ(size_t rank, size_t order,
   SEQUANT_ASSERT(rank == 1 &&
                  "sequant::mbpt::CC::tʼ(): only one-body perturbation "
                  "operator is supported now");
-  if (unitary()) {
-    SEQUANT_ASSERT(hbar_comm_rank_ &&
-                   "hbar_comm_rank must be specified for unitary ansatz");
-    SEQUANT_ASSERT(pertbar_comm_rank_ &&
-                   "pertbar_comm_rank must be specified for unitary "
-                   "ansatz");
-  }
+  if (unitary())
+    SEQUANT_ASSERT(pertbar_comm_rank_,
+                   "pertbar_comm_rank must be specified for unitary ansatz");
 
   // construct h1_bar
   // truncate h1_bar at rank 2 for one-body perturbation operator and at rank 4
@@ -379,10 +374,6 @@ std::vector<ExprPtr> CC::eom_r(nₚ np, nₕ nh) const {
     SEQUANT_ASSERT(
         get_default_context().spbasis() != SPBasis::Spinfree &&
         "spin-free basis does not yet support non particle-conserving cases");
-  if (unitary())
-    SEQUANT_ASSERT(hbar_comm_rank_ &&
-                   "hbar_comm_rank must be specified for unitary ansatz "
-                   "in CC::eom_r");
 
   // construct hbar
   const auto hbar = this->hbar();
--- a/SeQuant/domain/mbpt/models/cc.cpp
+++ b/SeQuant/domain/mbpt/models/cc.cpp
@@ -60,6 +60,12 @@ bool CC::unitary() const {
 
 std::optional<size_t> CC::hbar_comm_rank() const { return hbar_comm_rank_; }
 
+CC CC::with_hbar_comm_rank(size_t rank) const {
+  CC result(*this);
+  result.hbar_comm_rank_ = rank;
+  return result;
+}
+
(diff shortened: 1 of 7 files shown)


ajay-mk:
Weird error because I don't use Copilot for anything 🙂

ajay-mk:
This will need changes once #584 goes through.

SeQuant/domain/mbpt/op.hpp
```diff
@@ -1164,6 +1164,12 @@ ExprPtr F(bool use_tensor = true,
 /// A general operator of rank \p K
 ExprPtr θ(std::size_t K);
 
+/// @brief Makes a rank-\p rank number (replacement) operator
```
evaleev: **Naming — the item hardest to change later.**

`N` conventionally denotes the particle-number operator `Σ_p a†_p a_p`. This is a rank-*r* **replacement** operator (rank-1 is `Ê^p_q`), which is why the docs have to say "number (replacement) operator" everywhere. Two concrete costs:

- `CC::rdm` has to work around the collision with `CC::N` (the cluster rank) via an `op::` qualification plus an explanatory comment.
- `OpRegistry::validate_op` throws if a label is already registered, so adding `N` to *both* shipped registries breaks any downstream code that registers its own `N` — a very likely user label.

Would `E`, `Ê`, or `γ̂` work? Renaming after MPQC ships against this is much more expensive than now.

SeQuant/domain/mbpt/op.cpp
```diff
@@ -974,6 +974,30 @@ ExprPtr θ(std::size_t K) {
                   });
 }
 
+ExprPtr N(std::size_t rank) {
+  if (rank < 1) throw Exception("mbpt::op::N: rank must be >= 1");
```
evaleev: Diverges from its siblings on precondition handling: `θ`, `t`, `λ` all do `SEQUANT_ASSERT(K > 0)` *then* the registry check. Here it's a `throw` first, registry assert second.

Also `rank < 1` on a `std::size_t` is just `rank == 0` — worth spelling that way.

SeQuant/domain/mbpt/op.cpp
```diff
@@ -974,6 +974,30 @@ ExprPtr θ(std::size_t K) {
...
+  return ex<op_t>([]() -> std::wstring_view { return L"N"; },
+                  [rank]() -> ExprPtr {
+                    const auto space = get_complete_space(Spin::any);
+                    container::svector<Index> cres, anns;
+                    cres.reserve(rank);
+                    anns.reserve(rank);
+                    for (std::size_t i = 0; i < rank; ++i)
+                      cres.emplace_back(space, i + 1);
+                    for (std::size_t i = 0; i < rank; ++i)
+                      anns.emplace_back(space, rank + i + 1);
```
evaleev: **The free-index convention is the thing callers actually need, and it's currently undocumented.**

This emits creators `p_1…p_r` and annihilators `p_{r+1}…p_{2r}`, so `CC::rdm(2)` returns an expression whose ket indices are literally `p_3, p_4`. The docs (both here and in `cc.hpp`) write `q_1…q_r`, which corresponds to no label in the output — MPQC has to hard-code the real mapping. Please state the exact labels.

(Non-tmp ordinals are the right choice — they can't collide with the `Index::make_tmp_index` ordinals used by T/Λ. It's only the documentation that's missing.)

Nit: the two loops can be one.

SeQuant/domain/mbpt/models/cc.hpp
```diff
@@ -164,6 +169,29 @@ class CC {
...
+  /// reference expectation value of a similarity-transformed number operator.
+  /// Traditional ansatz:
+  /// \f$ \gamma^{p_1 \dots p_r}_{q_1 \dots q_r} = \langle 0| (1 +
+  /// \hat{\Lambda}) e^{-\hat{T}} \{ a^{p_1 \dots p_r}_{q_1 \dots q_r} \}
+  /// e^{\hat{T}} |0 \rangle \f$; unitary ansatz drops \f$ \hat{\Lambda} \f$ and
+  /// uses \f$ e^{-\hat{\sigma}} \dots e^{\hat{\sigma}} \f$ with \f$
+  /// \hat{\sigma} = \hat{T} - \hat{T}^\dagger \f$. The number-operator indices
+  /// are in the complete space, so they survive as the free indices of the
+  /// result. Returns the *linked* density (the number
+  /// operator is required to connect to the cluster amplitudes).
```
evaleev: Three doc issues in this block:

1. `\gamma^{p_1 \dots p_r}_{q_1 \dots q_r}` — there is no `q_*` in the output; the ket indices are `p_{r+1}…p_{2r}` (see my comment on `op.cpp`).
2. **The reference contribution is not included.** `ref_av` returns 0 for a bare normal-ordered operator, so this is the correlation part only and the caller adds `δ`. The `@return` line hints at it ("correlation part"), but the leading formula doesn't, and that's the part people read.
3. "Returns the *linked* density (the number operator is required to connect to the cluster amplitudes)" is true only for the traditional branch — the unitary branch passes an empty `OpConnections{}`. Consistent with `CC::energy`, but the sentence reads as unconditional.

Worth adding: the result is **not manifestly antisymmetric** under `p_1↔p_2` / `p_{r+1}↔p_{r+2}` for rank ≥ 2, since `N` carries no antisymmetrizer.

SeQuant/domain/mbpt/models/cc.cpp
```diff
@@ -198,6 +205,36 @@ std::vector<ExprPtr> CC::λ() const {
...
+
+  // 2. similarity transform N̄ = e^{-σ} N e^{σ} (σ = T, or T−T⁺ for unitary).
+  // Traditional ansatz: the expansion terminates exactly, so the default is the
+  // largest number of T's that can survive <0|(1+Λ) N̄|0>:
+  // Unitary ansatz: T⁺ contracts with T, the expansion never terminates, so
+  // there is no safe default; use the engine's hbar_comm_rank.
+  if (unitary())
+    SEQUANT_ASSERT(hbar_comm_rank_ &&
+                   "hbar_comm_rank must be specified for unitary ansatz in "
+                   "CC::rdm");
```
evaleev: This guard is dead code: the `CC` constructor already asserts `hbar_comm_rank_` for the unitary ansatz, and `with_hbar_comm_rank` only ever *sets* it — so it can never fire.

Worse, in assert-disabled builds *both* asserts vanish and the actual guard becomes `hbar_comm_rank_.value()` on the next line → `std::bad_optional_access`, which conflicts with the repo's preference for `sequant::Exception`. Either drop this, or make it a real `throw Exception(...)` that survives release builds.

Minor: `SEQUANT_ASSERT` now takes a message argument, so `SEQUANT_ASSERT(hbar_comm_rank_, "…")` beats the `cond && "msg"` idiom — though the rest of this file uses the old form.

SeQuant/domain/mbpt/models/cc.cpp
```diff
@@ -198,6 +205,36 @@ std::vector<ExprPtr> CC::λ() const {
...
+  const auto commutator_rank = comm_rank.value_or(
+      unitary() ? hbar_comm_rank_.value() : std::min(2 * rank, rank + N));
```
evaleev, line 229: The `min(2*rank, rank + N)` bound is the load-bearing correctness claim in this PR, and right now it's justified only in a commit message — nothing in CI checks it.

The argument checks out as I read it (each T must consume ≥1 of N's `2r` legs ⇒ `k ≤ 2r`; and T's `2k` quasi-creators must be absorbed by N's `2r` + Λ's `2N` legs ⇒ `k ≤ r + N`), but it deserves a test rather than a reader's re-derivation. See my comment on the test file.

SeQuant/domain/mbpt/models/cc.hpp
```diff
@@ -73,6 +73,11 @@ class CC {
   /// not set
   [[nodiscard]] std::optional<size_t> hbar_comm_rank() const;
 
+  /// @brief returns a copy of this object whose H̄ commutator truncation rank
+  /// is @p rank, leaving this engine unchanged; all other options are carried
+  /// over
+  [[nodiscard]] CC with_hbar_comm_rank(size_t rank) const;
```
evaleev, line 82: No validation that `rank >= 1`. `CC::λ` computes `hbar(commutator_rank - 1)` guarded only by a `SEQUANT_ASSERT`; with asserts disabled and `hbar_comm_rank_ == 0` that underflows to `SIZE_MAX`, and `mbpt::lst` then loops `for (k = 1; k <= SIZE_MAX; ++k)` — a hang.

This is reachable today via `Options{.hbar_comm_rank = 0}`, so it's pre-existing, but the new setter adds a second door and is the natural place to close it. A `SEQUANT_ASSERT(rank >= 1)` here (and ideally in the `Options` validation) is cheap.

Separately: this has no caller in the PR — `rdm` takes `comm_rank` directly. Fine as a general utility, just noting it's unrelated scope.

tests/unit/test_mbpt_cc.cpp
```diff
@@ -67,6 +67,40 @@ TEST_CASE("mbpt_cc", "[mbpt/cc][valgrind_skip]") {
...
+      REQUIRE(g);
+      REQUIRE(size(g) == 10);
+      auto G = CC{N}.rdm(2);
+      REQUIRE(G);
+      REQUIRE(size(G) == 94);
+      // an explicit comm_rank truncates early: cutting Γ at the 2nd nested
+      // commutator drops the terms the default (4th, exact) picks up
+      auto G2 = CC{N}.rdm(2, 2);
+      REQUIRE(G2);
+      REQUIRE(size(G2) == 71);  // 23 terms short of the exact 94
```
evaleev: These are term-count-only assertions, while the rest of this file uses `REQUIRE_THAT(..., EquivalentTo(L"..."))` for correctness (see the `λ` and `energy` sections). Counts drift silently with any canonicalization or `simplify` change and don't verify the equations at all — so correctness currently rests entirely on the MPQC cross-check, which this repo's CI can't see.

Highest-value additions, in order:

- **Prove the default `comm_rank` is exact.** One assertion turns the commit-message argument into something CI enforces:
  ```cpp
  REQUIRE_THAT(CC{2}.rdm(2), EquivalentTo(CC{2}.rdm(2, 6)));  // default is converged
  REQUIRE_THAT(CC{2}.rdm(1), EquivalentTo(CC{2}.rdm(1, 5)));
  ```
  Much stronger than `== 94`, and it's the claim most likely to be wrong.
- **Spell out the 1-RDM** with `EquivalentTo` — 10 terms is small enough, and it pins the free-index convention (`p_1`/`p_2`) into the suite.
- **Regression test for the bug fixed in the last commit**: `op::N` was unregistered and threw on printing. Nothing here exercises `to_latex(N(1))`.
- Missing: `op::N(0)` throw, UCC rank-2.

The `test_mbpt.cpp` addition is the right shape by contrast — it checks structure (`ncreators`, `nannihilators`, spaces) rather than counts.

evaleev (commented):
Inline version of the main points from my earlier summary comment. Design is sound overall — `op::N` deliberately bypassing `OpMaker` so its indices survive as γ's free indices is the right call, and the connectivity machinery in `vac_av.cpp` lines up correctly with an op that lowers to a bare `FNOperator`. The items below are ordered roughly by how hard they'd be to change after MPQC binds against this.

SeQuant/domain/mbpt/op.hpp
evaleev: Clarifying what this actually is, since it bears on the name: it's a **single replacer**, not the RDM operator `Σ γ^{i…}_{j…} a^{j…}_{i…}`.

The tensor-form generator returns `ex<FNOperator>(cre(cres), ann(anns))` and nothing else — no `Tensor` factor, no normalization, no summation — over fixed non-tmp ordinals, so the indices stay free. It is exactly `{ã^{p_1…p_r}_{p_{r+1}…p_{2r}}}`, one component.

The contrast with `θ` makes it sharp: `tensor::θ(K)` goes through `OpMaker::make`, which returns `apply_normalization(θ_tensor * NormalOperator(...))` over `make_tmp_index` indices, i.e. `1/(r!)² Σ θ^{…}_{…} ã^{…}_{…}`. **`op::N` is `θ` with the amplitude tensor and the normalization stripped off** — which is precisely why it has to bypass `OpMaker`.

So `CC::rdm(r)` returns the *element* γ^{p_1…p_r}_{p_{r+1}…p_{2r}} as an expression in free complete-space indices, not `Σ γ a`. That's the right primitive for the stated purpose (codegen of a γ tensor); the `Σ γ a` form would serve the opposite direction — evaluating expectation values against a density you already have. The construction is right; only the name is wrong.

### On the name

`E` is out, as noted elsewhere: SeQuant already prints the spin-free replacer as `ã`, and `E^p_q` is just the textbook name for that same object, so `E` would be a synonym for something that exists.

Out for the same class of reason: `γ̂`/`Γ̂` mislabel it as the density operator — the exact error `N` makes in the other direction.

Reserved labels are only `{Â, Ŝ, P̂, δ, s}`, so the field is otherwise open; the real constraints are avoiding index base keys (`p`, `a`, `i`, …) and existing registry entries (`h`, `f`, `f̃`, `g`, `θ`, `t`, `λ`, `R`, `L`, …).

Suggestion: make the label mirror the notation it *is* — registry label `L"ã"`, function `op::replacer(rank)` (or `op::ã`). Seeing `ã` at the operator level and `ã^{p_1}_{p_2}` in the lowered form is the same object twice, which is the point. It's unused, unreserved, and not an index key.

If a Greek single letter is preferred for symmetry with `θ`, `ν` is free and semantically uncommitted. I'd avoid `τ` — CC literature already uses it for the T₂+T₁T₁ combination.

evaleev (changes requested):
The physics and the construction are right — `op::N` bypassing `OpMaker` so its indices survive as γ's free indices is the correct primitive, the connectivity machinery lines up with an op that lowers to a bare `FNOperator`, and CI is green across all 16 checks. Details are in the inline comments; blocking items collected here.

**Blocking**

1. **The name `N`.** It denotes a single replacer `{ã^{p_1…p_r}_{p_{r+1}…p_{2r}}}`, not the number operator and not the density operator — and it squats a very likely user label in both shipped registries, where `OpRegistry::validate_op` turns a downstream re-registration into a throw. This is the piece that gets expensive once MPQC binds against it. See the thread on `op.hpp` for the reasoning and concrete alternatives.

2. **Document the free-index convention.** The result's ket indices are literally `p_{r+1}…p_{2r}`; the docs say `q_1…q_r`, which matches no label in the output. Callers can't use this without knowing the real mapping. Same block should state that the reference δ contribution is *not* included and that the result isn't manifestly antisymmetric for rank ≥ 2.

3. **Test that the default `comm_rank` is exact.** `min(2*rank, rank + N)` is the load-bearing correctness claim and is currently justified only in a commit message. `REQUIRE_THAT(CC{2}.rdm(2), EquivalentTo(CC{2}.rdm(2, 6)))` converts it into something CI enforces — much stronger than `size(G) == 94`, which drifts silently with any canonicalization change.

**Should fix, not blocking**

4. The unitary guard in `CC::rdm` is dead code (the ctor already asserts it), and in assert-disabled builds the real guard becomes `.value()` → `std::bad_optional_access`.
5. `with_hbar_comm_rank` doesn't validate `rank >= 1`; via `CC::λ`'s `commutator_rank - 1` that underflows to `SIZE_MAX` and hangs `mbpt::lst` in a release build.
6. `op::N` throws where `θ`/`t`/`λ` assert, with the registry check in the opposite order; `rank < 1` on an unsigned is `rank == 0`.
7. The "linked density" sentence is true only for the traditional branch.

Happy to re-review promptly — most of the list is small.

SeQuant/domain/mbpt/models/cc.hpp
ajay-mk: This is actually fine and `hbar_comm_rank = 0` is a valid use case, and only is a problem for `CC::λ`. `CC::λ` has an assertion to catch this case. The convention in the class is to use `SEQUANT_ASSERT` over `sequant::Exception`. 

~~`with_hbar_comm_rank` does not have a caller, but is used in MPQC for a non-RDM case. I can move this out of the PR is that is a blocker.~~ `with_hbar_comm_rank` has been dropped.

SeQuant/domain/mbpt/models/cc.cpp
ajay-mk: Agree with splitting the assertion, but `SEQUANT_ASSERT` is used throughout in the `CC` class and if we want to switch to `sequant::Exception` we should switch everything.

SeQuant/domain/mbpt/op.hpp
ajay-mk: Renamed to `ã`: registry label `L"ã"` in both shipped registries, and function `op::ã` in `mbpt` namespace. [31289a6](https://github.com/ValeevGroup/SeQuant/pull/578/commits/31289a6b611befae0e7f84df9163dd1062a3e029)

SeQuant/domain/mbpt/op.cpp
ajay-mk: Done: `SEQUANT_ASSERT(rank > 0, ...)` then the registry check, matching `θ`/`t`/`λ`. `rank < 1` is now `rank > 0`. [31289a6](https://github.com/ValeevGroup/SeQuant/pull/578/commits/31289a6b611befae0e7f84df9163dd1062a3e029)

SeQuant/domain/mbpt/op.cpp
ajay-mk: Stated on `op::ã` as a contract. Creators use ordinals `1 ... rank`, annihilators `rank+1 ... 2 * rank`. MPQC already relies on it for `closed_shell_spintrace` call.

Two loops are now unified. Two checks on the indices, not one: `to_latex(ã(2)->tensor_form())` in `test_mbpt.cpp` pins the operator's own ordinals, and at the RDM level the traditional Γ's free indices are pinned with `get_unique_indices`. [c3d1db0](https://github.com/ValeevGroup/SeQuant/pull/578/commits/c3d1db030000f0efb520557685f6ba4be334ab5c)

SeQuant/domain/mbpt/models/cc.hpp
ajay-mk: The formula reads `\gamma^{p_1 \dots p_r}_{p_{r+1} \dots p_{2r}}` over `\tilde{a}`, so it matches what is emitted. The missing reference contribution, the traditional-only scope of "linked", and the rank >= 2 antisymmetry caveat each got their own `@note`. 

Kept "not *manifestly* antisymmetric": the claim is that the form carries no antisymmetrizer, not that the object lacks the symmetry.

Also in this block: `@param comm_rank` referred to `\bar{N}`, which no longer names anything, and the `min(2*rank, rank + N)` argument moved out of the commit message into `cc.cpp`.

SeQuant/domain/mbpt/models/cc.cpp
ajay-mk: New tests added to verify this. [c3d1db0](https://github.com/ValeevGroup/SeQuant/pull/578/commits/c3d1db030000f0efb520557685f6ba4be334ab5c)

tests/unit/test_mbpt_cc.cpp
ajay-mk: Both 1-RDMs are now spelled out with `EquivalentTo`; the `to_latex` regression, the rank-0 precondition, and the unitary rank-2 density are now tested.

For the 2-RDMs the *expressions* stay term counts (94, 71, 24) — 94 terms is not worth spelling out. The rank-2 path is not count-only though: `TCC RDM default comm_rank` compares `rdm(2)` against `rdm(2, 6)` as full expressions, and the traditional Γ has its free indices pinned with `get_unique_indices` (ket `{p_1, p_2}`, bra `{p_3, p_4}`, no aux). [c3d1db0](https://github.com/ValeevGroup/SeQuant/pull/578/commits/c3d1db030000f0efb520557685f6ba4be334ab5c)

SeQuant/domain/mbpt/models/cc.cpp
ajay-mk: Dropped, per the first of your two options. [5743c9e](https://github.com/ValeevGroup/SeQuant/pull/578/commits/5743c9ecf0ee90a0272d0eca5b2dce25404316b6)

It was dead in three places rather than one — `tʼ` and `eom_r` carried the same copy — so the constructor is now the sole guard. `tʼ` keeps its `pertbar_comm_rank` check, which the constructor does not cover. Message-argument form used on the lines this touches.

evaleev (approved):

PR #581, opened 2026-07-27
Bernoulli Expansion based Unitary Coupled-Cluster

ajay-mk:
Adds the **Bernoulli expansion** of the UCC similarity-transformed Hamiltonian $\bar{H} = e^{-\sigma} H e^{\sigma}$, $\sigma = T - T^\dagger$, as an alternative to the BCH expansion in `mbpt::CC`. Eq. (45) defines the sum, and Eqs. (46)–(50) define $\bar H^0$ – $\bar H^4$ in [Liu, Asthana, Cheng & Mukherjee, *JCP* **148**, 244110 (2018)](https://doi.org/10.1063/1.5030344)

Companion PR: [MPQC4 #792](https://github.com/ValeevGroup/mpqc4/pull/792).

## Changes
- Adds the Bernoulli-number expansion of the UCC similarity-transformed Hamiltonian $\bar H = e^{-\sigma}He^{\sigma}$ as an alternative to BCH in `mbpt::CC`, plus per-block $\bar H$ truncation for block-truncated EOM (qUCCSD and its IP/EA analogues).
- `CC::Options::hbar_expansion`: Selects `BCH` or `Bernoulli`; the former is the default. Bernoulli requires
`Ansatz::U` and `hbar_comm_rank`.
- New Bernoulli expansion related logic:
	- `bernoulli::hbar(N, rank, skip1)` assembles $\bar H^0$ – $\bar H^4$ (Eqs. (46)–(50) of Ref. 1) rank by rank. $H$ splits as $F+V$, and every operator splits into its pure excitation/de-excitation part ($O_N$) and the rest ($O_R$).
	- `detail::wick_reduce` — Wick-reduces while keeping partial contractions
  (needed so nesting produces operators, not scalars). Must run with `use_topology(false)`: its weight bookkeeping is only valid on the fully-contracted path, and it defaults to on.
	- `detail::N_part` / `detail::R_part` — the $O_N$ / $O_R$ split. Only $N$ is block-resolved; $R$ stays in compact general-index form, which keeps the nested commutators that consume it cheap.
	- `bernoulli::hbar()` returns a **tensor-level** expression (BCH's `hbar()` returns an operator-level one). Callers have to take care of this.
- `CC::eom_r`: per-block H̄ truncation (`block_ranks`):
	- New optional argument: a row-major $K\times K$ matrix, one $\bar H$ truncation rank per block of the EOM secular matrix instead of one uniform $\bar H$.
	-  Manifolds are ordered by ascending rank, so the same matrix serves EE, IP and EA. qUCCSD is `{2,1,1,0}`. See [Liu & Cheng, *JCP* **155**, 174102 (2021)](https://doi.org/10.1063/5.0062090), Sec. II C; [Zhang & Liu, *JCTC* **22**, 3341 (2026)](https://doi.org/10.1021/acs.jctc.5c01991), Fig. 1, for the IP/EA analogues.
	- Each block is the sandwich ⟨i|H¯|j⟩ with its diagonal scalar taken from the same H¯ truncation, not the commutator form. Under Bernoulli, each block's H¯ has its N part (the ground-state amplitude residual) removed. The removed terms vanish at converged amplitudes when the block rank equals `hbar_comm_rank`; below that rank, the values change too.

## Tests

- `tests/unit/test_mbpt_cc.cpp`: Wick reduction, N/R split, $\bar H$ structure, config validation, block-truncated EOM term counts and rejection cases.
- `tests/integration/ucc.cpp`: BCH/Bernoulli term-count pins at ranks 2–4.
- Implementation is validated numerically against literature, see [MPQC4 #792](https://github.com/ValeevGroup/mpqc4/pull/792) for details.

## qUCCSD Example

```c++
#include <SeQuant/core/context.hpp>
#include <SeQuant/core/expr.hpp>
#include <SeQuant/core/op.hpp>
#include <SeQuant/core/tensor_canonicalizer.hpp>
#include <SeQuant/domain/mbpt/context.hpp>
#include <SeQuant/domain/mbpt/convention.hpp>
#include <SeQuant/domain/mbpt/models/cc.hpp>
#include <SeQuant/domain/mbpt/op.hpp>

#include <iostream>

using namespace sequant;
using namespace sequant::mbpt;

int main() {
  set_locale();
  set_default_context({.index_space_registry_shared_ptr = make_sr_spaces(),
                       .vacuum = Vacuum::SingleProduct,
                       .metric = IndexSpaceMetric::Unit,
                       .spbasis = SPBasis::Spinor});
  TensorCanonicalizer::set_cardinal_tensor_labels(cardinal_tensor_labels());
  set_default_mbpt_context({.op_registry_ptr = make_legacy_registry()});

  const CC cc(2, {.ansatz = CC::Ansatz::U,
                  .hbar_comm_rank = 2,
                  .hbar_expansion = CC::HbarExpansion::Bernoulli});


  // qUCCSD: Bernoulli order 3 for the energy, 2 for ground-state amplitudes; [2,1,1,0] for excited states

  // energy
  const auto energy = cc.energy(3);
  std::wcout << "E        : " << energy->size() << " terms\n";

  // t amplitudes
  const auto t = cc.t();
  for (std::size_t p = 1; p < t.size(); ++p)
    std::wcout << "residual" << p << ": " << t[p]->size() << " terms\n";

  // EE
  const auto sigma = cc.eom_r(nₚ(2), nₕ(2), {2, 1, 1, 0});
  for (std::size_t p = 1; p < sigma.size(); ++p)
    std::wcout << "sigma" << p << "   : " << sigma[p]->size() << " terms\n";

  // IP/EA use the same block ranks, with R of unequal particle/hole rank
  std::wcout << "IP sigma1: " << cc.eom_r(nₚ(1), nₕ(2), {2, 1, 1, 0})[0]->size()
             << " terms\n";
}

```

## Important References:
- [Liu, Asthana, Cheng & Mukherjee, *JCP* **148**, 244110 (2018)](https://doi.org/10.1063/1.5030344): Bernoulli expansion and qUCCSD equations
- [J. Liu, D. A. Matthews, and L. Cheng, *JCTC* **18**, 2281 (2022)](https://doi.org/10.1021/acs.jctc.1c01210): Efficient qUCCSD implementation and higher-order methods
- [Liu & Cheng, *JCP* **155**, 174102 (2021)](https://doi.org/10.1063/5.0062090): Reference EE results
- [Zhang & Liu, *JCTC* **22**, 3341 (2026)](https://doi.org/10.1021/acs.jctc.5c01991): Reference IP/EA results

### DIFF
--- a/SeQuant/domain/mbpt/bernoulli.cpp
+++ b/SeQuant/domain/mbpt/bernoulli.cpp
@@ -10,6 +10,7 @@
 #include <SeQuant/core/utility/indices.hpp>
 #include <SeQuant/core/utility/macros.hpp>
 #include <SeQuant/core/wick.hpp>
+#include <SeQuant/domain/mbpt/context.hpp>
 #include <SeQuant/domain/mbpt/op.hpp>
 
 #include <range/v3/algorithm/all_of.hpp>
@@ -39,8 +40,8 @@
 
 namespace {
 
-/// Returns the single residual fermionic NormalOperator carried by @p term, or
-/// nullptr when it has none (a pure scalar / fully-contracted term).
+/// Returns the residual fermionic NormalOperator in @p term, if any.
+/// @throws Exception if @p term contains more than one.
 const sequant::NormalOperator<sequant::Statistics::FermiDirac>* find_nop(
     const sequant::ExprPtr& term) {
   using namespace sequant;
@@ -51,9 +52,8 @@ const sequant::NormalOperator<sequant::Statistics::FermiDirac>* find_nop(
     const NormalOperator<Statistics::FermiDirac>* found = nullptr;
     for (const auto& f : term.as<Product>().factors())
       if (f.is<NormalOperator<Statistics::FermiDirac>>()) {
-        SEQUANT_ASSERT(!found,
-                       "find_nop: term carries >1 NormalOperator; wick_reduce "
-                       "is expected to leave at most one residual operator");
+        if (found)
+          throw Exception("find_nop: term contains multiple NormalOperators");
         found = &f.as<NormalOperator<Statistics::FermiDirac>>();
       }
 
@@ -263,6 +263,8 @@ ExprPtr R_part(const ExprPtr& expr, std::size_t cutoff, std::size_t min_rank) {
 // Each H̄^k below transcribes its equation. A subscript R/N means "take that
 // part of the commutator before the next nesting".
 ExprPtr hbar(std::size_t N, std::size_t rank, bool skip1) {
+  if (get_default_mbpt_context().csv() == CSV::Yes)
+    throw Exception("bernoulli::hbar: CSV is not supported");
   if (rank > 4)
     throw Exception("bernoulli::hbar: only ranks [0,4] are implemented");
 
--- a/SeQuant/domain/mbpt/bernoulli.cpp
+++ b/SeQuant/domain/mbpt/bernoulli.cpp
@@ -117,7 +117,8 @@ namespace sequant::mbpt::bernoulli {
 
 namespace detail {
 
-ExprPtr wick_reduce(ExprPtr expr) {
+ExprPtr wick_reduce(const ExprPtr& expr_in) {
+  auto expr = expr_in->clone();
   simplify(expr);
   FWickTheorem wick{expr};
   // use_topology defaults to ON, so it must be turned off explicitly. It keeps
@@ -266,7 +267,7 @@ ExprPtr R_part_reduced(const ExprPtr& reduced, std::size_t cutoff) {
 /// header): after expansion every residual index is definite, so the N/R
 /// classifier can act on it.
 ExprPtr expand_to_blocks(const ExprPtr& expr_in) {
-  return expand_to_blocks_reduced(wick_reduce(expr_in->clone()));
+  return expand_to_blocks_reduced(wick_reduce(expr_in));
 }
 
 /// N part of @p expr at truncation @p cutoff (see header): block-resolve, then
@@ -282,7 +283,7 @@ ExprPtr N_part(const ExprPtr& expr, std::size_t cutoff) {
 /// equals the fully block-resolved remainder, and the compact expr makes the
 /// nested commutators that consume R operate on far fewer terms.
 ExprPtr R_part(const ExprPtr& expr, std::size_t cutoff) {
-  auto reduced = wick_reduce(expr->clone());
+  auto reduced = wick_reduce(expr);
   return R_part_reduced(reduced, cutoff);
 }
 
--- a/SeQuant/domain/mbpt/bernoulli.cpp
+++ b/SeQuant/domain/mbpt/bernoulli.cpp
@@ -4,29 +4,38 @@
 #include <SeQuant/core/index.hpp>
 #include <SeQuant/core/index_space_registry.hpp>
 #include <SeQuant/core/op.hpp>
+#include <SeQuant/core/rational.hpp>
+#include <SeQuant/core/utility/exception.hpp>
 #include <SeQuant/core/utility/expr.hpp>
 #include <SeQuant/core/utility/indices.hpp>
 #include <SeQuant/core/utility/macros.hpp>
 #include <SeQuant/core/wick.hpp>
+#include <SeQuant/domain/mbpt/op.hpp>
 
 #include <range/v3/algorithm/all_of.hpp>
 #include <range/v3/algorithm/any_of.hpp>
 #include <range/v3/algorithm/none_of.hpp>
+
+#include <string>
+#include <utility>
 
 // Bernoulli expansion of the unitary-CC similarity-transformed Hamiltonian
 // H̄ = e^{−σ} H e^{σ}, σ = T − T† (anti-Hermitian). Because σ mixes excitation
 // and de-excitation the plain BCH series does not terminate; the Bernoulli
 // expansion rewrites it so that Bernoulli numbers are the expansion
 // coefficients, leaving the final truncation at a chosen commutator rank as the
-// only approximation. H is split as F (Fock, rank-preserving) + V
-// (fluctuation potential), and every operator O is split into O_N (all
-// excitation and de-excitation operators) and O_R = O − O_N.
+// only approximation. H is split as F (Fock, rank-preserving) + V (fluctuation
+// potential), and every operator O is split into O_N (all excitation and
+// de-excitation operators) and O_R = O − O_N. At a converged RHF/UHF reference
+// two cancellations hold: F survives only in H̄¹, and the higher orders carry
+// only R-subscripted inner commutators.
 //
-// This file builds that expansion in three layers: an operator-valued Wick
-// reduction, the N/R operator split (here), and the rank-by-rank assembly of
-// H̄. All equation references are to 10.1063/1.5030344 (Sec. III B); the N/R
-// split and the UCC amplitude condition V̄_N = 0 are defined above and at
-// Eq. (43).
+// All equation references are to 10.1063/1.5030344 (Sec. III B): superoperator
+// inversion Eqs. (36)-(39); Bernoulli numbers B₁=−1/2, B₂=1/12, B₃=0, B₄=−1/720
+// Eq. (40); the N/R split and the UCC amplitude condition V̄_N = 0 above and at
+// Eq. (43); the iterative recursion for V̄ Eq. (44); the rank-by-rank operators
+// H̄⁰..H̄⁴ Eqs. (45)-(50). Cancellation #1 (F enters only H̄¹) is stated just
+// below Eq. (50).
 
 namespace {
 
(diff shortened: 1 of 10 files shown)


SeQuant/domain/mbpt/bernoulli.cpp
```diff
@@ -0,0 +1,401 @@
...
+/// Assembles H̄ order by order (see header), summing H̄⁰..H̄^rank of Eq. (45).
+/// Each H̄^k below is a direct transcription of its equation. A subscript R/N on
+/// a commutator means "form the commutator, then keep only its R/N part before
+/// the next nesting".
+ExprPtr hbar(std::size_t N, std::size_t rank, bool skip1) {
+  if (rank > 4)
+    throw Exception("bernoulli::hbar: only ranks [0,4] are implemented");
+
+  using namespace detail;
+  const auto cutoff = N;
```
evaleev, line 275: **high — the N/R cutoff ignores `skip1`.**

`const auto cutoff = N;` does not account for skipped singles. The N/R split's own doc (lines 81-85) ties the cutoff to "σ here is truncated at rank N", and the whole resummation rests on the amplitude condition V̄_N = 0. With `skip1 == true`, σ = T − T† carries ranks 2..N, not 1..N, so a rank-1 pure excitation/de-excitation carries no amplitude and `<Φ_i^a|H̄|0>` is *not* zero — yet `is_N_term` still classifies those terms as N, since it only checks an upper bound (`std::max(ncre, nann) > cutoff`, line 95).

Reachable through the public API and accepted by the constructor (only `ansatz_ == U` and `hbar_comm_rank_` are asserted):

```cpp
CC cc(2, {.ansatz = CC::Ansatz::U, .skip_singles = true,
          .hbar_comm_rank = 2, .hbar_expansion = CC::HbarExpansion::Bernoulli});
cc.t();
```

The singles-rank terms are then weighted 1/12, −1/720, … (N chains) instead of ¼, ⅛, … (R chains), so the UCCD equations come out silently wrong. Either lower-bound the cutoff (`skip1 ? 2 : 1`) or have `bernoulli::hbar` reject `skip1`.

SeQuant/domain/mbpt/models/cc.cpp
```diff
@@ -364,13 +401,104 @@ namespace {
...
+    const auto bra = bra_of(bp, bh);
+    auto acc = std::make_shared<Sum>();
+    for (size_t j = 0; j < K; ++j) {
+      const auto [kp, kh] = manifolds[j];
+      const auto& hbar_ij = hbars.at(ranks.at(i * K + j));
+      const auto ket = ket_of(kp, kh);
+      acc->append(vev(bra * hbar_ij * ket));
+      // -<0|H̄^(k_ii)|0>, written as <i|r_i H̄|0> so Wick keeps E's summed
+      // indices disjoint from the block's external ones.
+      if (i == j) acc->append(ex<Constant>(-1) * vev(bra * ket * hbar_ij));
```
evaleev, line 488: **medium — each diagonal block subtracts a different energy zero.**

The shift uses `hbar_ij`, i.e. `<0|H̄^(k_ii)|0>` at *that block's* truncation rank, so different diagonal blocks are referenced to different energy zeros.

Concretely, with the qUCCSD ranks used in the new test — `cc.eom_r(nₚ(2), nₕ(2), {2,1,1,0})` — the SS block subtracts `<0|H̄⁰+H̄¹+H̄²|0>` (the nonzero rank-2 correlation energy; the new `bernoulli_hbar_structure` test pins `E2_contrib ≠ 0`), while the DD block subtracts `<0|F+V|0> = 0`, since those operators are normal-ordered w.r.t. the reference. The returned object is then not `(H̄ − E·I)R̂` as the header `@note` claims, and the 2h2p eigenvalues are offset from the 1h1p ones by the full correlation energy (~0.1-0.2 Ha, i.e. several eV).

Eq. (10) of 10.1063/5.0062090 splits off a single `E_gr`. Could you confirm against the paper whether the per-block energy is intended here?

SeQuant/domain/mbpt/models/cc.cpp
```diff
@@ -364,13 +401,104 @@ namespace {
...
+  SEQUANT_ASSERT(np > 0 || nh > 0, "Unsupported excitation order");
   if (np != nh)
     SEQUANT_ASSERT(
-        get_default_context().spbasis() != SPBasis::Spinfree &&
+        get_default_context().spbasis() != SPBasis::Spinfree,
         "spin-free basis does not yet support non particle-conserving cases");
 
+  // Bernoulli always takes the blocked path: the uniform one below commutes H̄
+  // with an operator-level R, which a tensor-level H̄ cannot take part in. An
+  // empty matrix there means uniform truncation at hbar_comm_rank.
```
evaleev: **low — on the BCH path, a uniform `block_ranks` is not equivalent to an empty one.**

The header doc says "Empty (the default) selects the uniform H̄ at `hbar_comm_rank` everywhere", but for `HbarExpansion::BCH` the two paths differ. `eom_r(np, nh)` on the unitary path builds `<i|[H̄,R]|0>`, which subtracts `<i|r_j H̄|0>` for *every* j; `eom_r_blocked` subtracts it only for `j == i`. For `j < i` that term is nonzero off-shell (the leftover bra operators contract into H̄'s rank-(i−j) amplitude residual), and BCH does not get the N-part removal that kills them on the Bernoulli path.

```cpp
CC bch(2, {.ansatz = U, .hbar_comm_rank = 2});
bch.eom_r(nₚ(2), nₕ(2), {2,2,2,2});  // differs from bch.eom_r(nₚ(2), nₕ(2))
```

by disconnected `r_1 × (doubles residual)` terms. The new test only pins the equivalence for `K == 1` (`nₚ(1), nₕ(1), {2}`), where the two paths trivially coincide.

SeQuant/domain/mbpt/models/cc.cpp
```diff
@@ -364,13 +401,104 @@ namespace {
...
+    if (rp == 0 || rh == 0) break;
+  }
+
+  std::ranges::reverse(manifolds);
+  const auto K = manifolds.size();
+  // empty means uniform truncation at hbar_comm_rank in every block
+  const std::vector<size_t> ranks =
+      block_ranks.empty() ? std::vector<size_t>(K * K, hbar_comm_rank().value())
+                          : block_ranks;
+  SEQUANT_ASSERT(ranks.size() == K * K,
```
evaleev: **low — this precondition disappears in release builds.**

`SEQUANT_ASSERT` expands to `do {} while(0)` unless `SEQUANT_ASSERT_ENABLED` is defined (`SeQuant/core/utility/macros.hpp:131`), so the `ranks.size() == K*K` check is gone in a release build. An oversized `block_ranks` (e.g. a 3×3 matrix passed for a 2-manifold IP problem) is then silently accepted, and only its first `K*K` entries are read via `ranks.at(i*K+j)` — wrong truncation ranks with no diagnostic.

Same applies to `SEQUANT_ASSERT(unitary(), ...)` at line 414: a non-unitary `CC` with `block_ranks` proceeds into the blocked path in release instead of erroring. Both look like they want `throw Exception(...)`.

SeQuant/domain/mbpt/bernoulli.cpp
```diff
@@ -0,0 +1,401 @@
...
+    return &term.as<NormalOperator<Statistics::FermiDirac>>();
+
+  if (term.is<Product>()) {
+    const NormalOperator<Statistics::FermiDirac>* found = nullptr;
+    for (const auto& f : term.as<Product>().factors())
+      if (f.is<NormalOperator<Statistics::FermiDirac>>()) {
+        // The one-residual-operator invariant is load-bearing: N/R
+        // classification reads this operator alone, so a second one would be
+        // silently ignored and misclassify the term.
+        SEQUANT_ASSERT(!found,
```
evaleev: **low — the load-bearing invariant is release-only advisory, and `NormalOperatorSequence` slips through.**

Two things here:

1. The "one residual `NormalOperator`" invariant this comment documents as load-bearing is enforced only by `SEQUANT_ASSERT`, a no-op without `SEQUANT_ASSERT_ENABLED`. In release the loop silently keeps the *last* match and `is_N_term` misclassifies the term.
2. `find_nop` recognizes only `NormalOperator` factors. A `NormalOperatorSequence` factor is *not* a `NormalOperator`, so it returns `nullptr`, the term is treated as a fully-contracted scalar and unconditionally routed to R — and no assert fires at all. Worth handling `NormalOperatorSequence` explicitly rather than falling through.

SeQuant/domain/mbpt/bernoulli.cpp
```diff
@@ -0,0 +1,401 @@
...
+  return result;
+}
+
+ExprPtr wick_commutator(const ExprPtr& A, const ExprPtr& B) {
+  // A and B are built independently, so shared labels (both a block-resolved
+  // part and sigma carry a/i) would fuse two independent summations in A*B.
+  // Reindex B to fresh temporaries; canonicalization restores tidy labels.
+  container::map<Index, Index> repl;
+  for (const auto& idx : get_used_indices(B))
+    repl.emplace(idx, Index::make_tmp_index(idx.space()));
```
evaleev, line 135: **low — `make_tmp_index` drops proto-indices, breaking CSV.**

`Index::make_tmp_index(idx.space())` discards proto-indices. `get_used_indices` returns CSV indices with their protos intact (`a_1<i_1,i_2>`) and `transform_indices` matches them exactly, so under `mbpt::CSV::Yes` this replaces `a_1<i_1,i_2>` with a bare `a_100` and destroys the cluster-specific structure. Nothing in `bernoulli::hbar` or `CC` rejects a CSV context.

Either use `make_tmp_index(idx.space(), idx.proto_indices())` or add an explicit precondition check. Line 215 in `expand_to_blocks_reduced` has the same shape, though there the expanded indices come from the Hamiltonian and carry no protos.

SeQuant/domain/mbpt/models/cc.cpp
```diff
@@ -69,13 +81,22 @@ bool CC::use_topology() const { return use_topology_; }
...
   // is only equivalent to the commutator once the caller supplies operator
   // connectivity to ref_av (see lst_options() and the @warning on hbar())
   return mbpt::lst(H(), T(N, skip_singles()), truncation, lst_options());
 }
 
 ExprPtr CC::energy(std::optional<size_t> comm_rank) const {
+  // Bernoulli: the hbar expansion is at tensor level, call the tensor level
+  // ref_av directly. No connectivity or screening.
+  if (hbar_expansion_ == HbarExpansion::Bernoulli) {
+    const auto erank = comm_rank.value_or(hbar_comm_rank_.value());
```
evaleev, line 92: **low — `bad_optional_access` instead of a configuration error in release.**

`hbar_comm_rank_.value()` (also at line 427) relies on the constructor's `SEQUANT_ASSERT(hbar_comm_rank_, ...)`, which vanishes in release. A `CC` built with `.hbar_expansion = Bernoulli` and no `hbar_comm_rank` then throws `std::bad_optional_access` out of `energy()`/`eom_r()` rather than reporting the actual configuration error. `value_or` plus an explicit `throw Exception(...)` would match the project's `sequant::Exception` preference.

evaleev (commented):
Review pass over the whole PR (9 files, ~1.3k diff lines): the new `bernoulli.{cpp,hpp}`, the `cc.{cpp,hpp}` changes, and the supporting machinery they lean on.

**Checked and found correct**, for the record: the manifold enumeration in `eom_r_blocked` mirrors `op::R`'s loop exactly and the result-index map `min(bp,bh)` is in range and collision-free for EE/IP/EA; the `-E` diagonal shift written as `<i|r_i H̄|0>` really does factor into `E·<i|r_i|0>` (the ket operators are all right-type and can only contract into the bra), and its absence off-diagonal is right; the `nest` memo keys are unambiguous and the `memo`/`hbars` iterator use is safe despite `container::map` being a `flat_map`; `operator*`/`operator+`/`Product::append` deep-copy, so memoized subexpressions survive later `simplify` calls; `Product` hashing excludes the scalar, so `HashingAccumulator` merges the `add()` wrappers correctly; and `NormalOperator` is an `AbstractTensor`, so `get_used_indices`/`transform_expr` reach its indices and TensorNetwork canonicalization preserves non-c-number ordering (the commutator does not collapse).

**Findings** are inline. Summary:

1. **high** — `bernoulli.cpp`: the N/R cutoff ignores `skip1`, so `skip_singles` + Bernoulli silently produces wrong UCCD equations.
2. **medium** — `cc.cpp`: the diagonal shift uses each block's own truncation rank, so different diagonal blocks are referenced to different energy zeros.
3. **low** — `cc.cpp`: on the BCH path, a uniform `block_ranks` matrix is not equivalent to an empty one, contrary to the header doc.
4-7. **low** — four preconditions expressed as `SEQUANT_ASSERT`, which compiles to `do {} while(0)` without `SEQUANT_ASSERT_ENABLED`; plus a `find_nop` gap on `NormalOperatorSequence` and a dropped-proto-index issue under CSV.

SeQuant/domain/mbpt/models/cc.hpp
```diff
@@ -54,6 +61,12 @@ class CC {
     /// perturbation operator; must be specified if unitary ansatz is used in
     /// perturbed amplitude derivation
     std::optional<size_t> pertbar_comm_rank = std::nullopt;
+    /// choice of H̄ expansion; Bernoulli requires a unitary ansatz.
+    /// @note the Bernoulli H̄ is assembled at the tensor level and does not go
+    /// through CC::ref_av(), which is what forwards `screen` and
+    /// `use_topology`; it calls `op::tensor::ref_av()` with that function's own
+    /// defaults instead
+    HbarExpansion hbar_expansion = HbarExpansion::BCH;
```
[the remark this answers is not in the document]
ajay-mk: The Bernoulli implementation requires `Ansatz::U`. The option documentation now states that explicitly. Fixed in [`61ecc14dc`](https://github.com/ValeevGroup/SeQuant/commit/61ecc14dcd35cc048d4b9eb9226d5f5873bbc661).

SeQuant/domain/mbpt/bernoulli.cpp
ajay-mk: Real bug, fixed in [`bc7be1ee2`](https://github.com/ValeevGroup/SeQuant/commit/bc7be1ee24a983d82753e9de817cc050bfd2d195).

SeQuant/domain/mbpt/models/cc.cpp
ajay-mk: On the BCH path, an empty matrix builds the commutator form, while an explicit uniform matrix builds each block. Under Bernoulli, both requests dispatch to the blocked implementation, so empty and uniform matrices are the same request. The API documentation now makes that distinction, and the Bernoulli test checks both returned manifolds rather than only the singles element. Fixed in [`935efac0d`](https://github.com/ValeevGroup/SeQuant/commit/935efac0d45d4380944ce616357af547ed45a99c).

SeQuant/domain/mbpt/models/cc.cpp
ajay-mk: Switched to an `Exception` in [`935efac0d`](https://github.com/ValeevGroup/SeQuant/commit/935efac0d45d4380944ce616357af547ed45a99c). Additional guards also added.

SeQuant/domain/mbpt/bernoulli.cpp
ajay-mk: In the current Bernoulli path `find_nop` will not encounter a`NormalOperatorSequence`. `find_nop` now throws if a term contains multiple residual `NormalOperator` factors. Addressed in [`b075efeda`](https://github.com/ValeevGroup/SeQuant/commit/b075efeda689f66b9de98db7db45dd6c67aeb7a2).

SeQuant/domain/mbpt/models/cc.cpp
ajay-mk: Reviews keep bringing this up. The `CC` class constructor now throws when a unitary ansatz has no `hbar_comm_rank`. It also rejects Bernoulli with an ansatz other than `Ansatz::U`. Addressed in [`b075efeda`](https://github.com/ValeevGroup/SeQuant/commit/b075efeda689f66b9de98db7db45dd6c67aeb7a2).

SeQuant/domain/mbpt/bernoulli.cpp
ajay-mk: After some investigation with Codex, CSV + Bernoulli expansion needs more support and is outside this PR's scope. For now I have disabled it, can be addressed later.

SeQuant/domain/mbpt/models/cc.cpp
ajay-mk: Eq. (10) separates E_gr from the untruncated Hbar before Sec. II C applies a different truncation to each block. In the implementation, each block is built from its own Hbar^(k), and its diagonal removes the fully contracted scalar from that same Hbar^(k). This is why the qUCCSD DD block subtracts zero.

This can be verified in two ways:

- Symbolically, the DD block contains exactly the five bare-integral terms specified by Eqs. (30), (34), (38), and (48), and the implementation reproduces all five.
- Numerically, the MPQC-side PR reproduces the published qUCCSD excitation energies within meV. The paper reports the energies only to two decimal places in eV. If the E_gr separation were wrong in the way suggested, the much larger errors would be visible.

Documentation updated in [`935efac0d`](https://github.com/ValeevGroup/SeQuant/commit/935efac0d45d4380944ce616357af547ed45a99c).

ajay-mk:
This is now good to go. All review comments have been addressed:
- Updated the documentation to state that the Bernoulli expansion requires Ansatz::U. Fixed in 61ecc14dc. 
- Fixed the N/R split when singles are skipped by tracking the lowest rank carried by sigma. Fixed in bc7be1ee2.
- Verified the per-block E_gr subtraction symbolically against Eqs. (30), (34), (38), and (48), and numerically against the published qUCCSD excitation energies. Documentation updated in 935efac0d.
- Clarified the difference between empty and uniform block_ranks on the BCH path in 935efac0d.
- Changed the Unitary-ansatz and K x K block_ranks checks to `sequant::Exception `validation.
- `find_nop()` now throws if a term contains multiple residual NormalOperator factors.
- CSV with the Bernoulli expansion requires broader support, so bernoulli::hbar now rejects a CSV context instead of silently producing an incorrect expression. Fixed in b075efeda.
- The CC constructor now throws when a unitary ansatz has no hbar_comm_rank and rejects Bernoulli with an ansatz other than Ansatz::U. Fixed in b075efeda.
- Added follow-up Bernoulli and EOM documentation corrections in 5f18becfa.

All tests in SeQuant and MPQC still pass.

evaleev (approved):

PR #583, opened 2026-07-29
Multimode batched evaluation of factorized CC (cost model + placement + dry-run)

evaleev:
# Multimode batched evaluation of factorized coupled-cluster equations

Adds cost-model-driven **multimode batching** to the SeQuant evaluator so
large-system CSV/PNO-CC residuals can be evaluated without forming their
largest transients whole. Six squashed commits (dry-run backend, optimizer,
evaluator, supporting core, tests, docs).

## What it does

- **Batch modes** with two kinds: *external* (occ / PNO pair, free on the result
  -> scattered into disjoint slices) and *contracted* (DF aux, summed ->
  accumulated). Loops nest external-outside-contracted.
- **Perf-first cost model** (`DenseTimeSpace`): minimizes flops with
  `peak_threshold` as a ceiling; role-split (contracted/external) batchability;
  order-aware placement over the combined nest.
- **Batched evaluator**: cache scope chain with fall-through; **slice-on-use**
  (a cached intermediate fetched from an outer scope is sliced to the current
  block) decouples correctness from placement; **per-level placement** driven by
  a per-canonical **lifetime mask** (cross-occurrence proto-aware meet) unioned
  with contracted residency; iterative (stack-safe) tree traversal.
- **Dry-run cost-profile** backend predicts peak/flops/exec over the same IR.

## C60 PNO-CCSD dry run (55-term residual, aux K@256, occ@8, 100 GB budget)

| batching | dry-run peak | schedule flops (DP cost) | modeled roofline time |
|---|---:|---|---:|
| none | 38 897 GB | baseline | 1.06e17 |
| contracted-aux only | 6 047 GB (**6.4x**) | **identical schedule** | 2.94e16 |
| ext-occ + contracted-aux | 443.6 GB (**87.7x**) | **identical schedule** | 1.13e17 |

The DP selects the **same factorization** regardless of what is batchable
(flops are unchanged); batching only slices modes to lower the **peak**. The
roofline-time column moves because a giant intermediate executed whole is
memory-bound (`machine_balance x traffic`) but compute-bound when sliced -- the
cache-blocking win of the *same* schedule, not a cheaper one. Recompute overhead
(`avoidable_time`) is 1.8% -> 6.5% -> 39.8% as slicing gets more aggressive.

## Validation

- Units: `[eval]` 449, `[lifetime_mask]` 76, `[optimize]` 628 assertions green;
  OFF (order-blind) path byte-identical.
- MPQC he10 CSV-CCk on this stack: batched (371 external scatter + 962 aux group
  events) matches unbatched to < 1e-9, within the 1e-7 precision, no aborts.

Follow-ups (non-blocking, from the final review): dedup the proto-expansion
helper; add a real-forest hidden-tag hash-regression test; revisit the
`stamp_lifetime_masks` `const_cast`.


evaleev:
Superseded by https://github.com/ValeevGroup/SeQuant/pull/613 (same work, history squashed and legacy routes removed).

PR #584, opened 2026-07-30
Change defaults for `mbpt::lst`

ajay-mk:
`mbpt::lst`'s default assumed the caller knew about operator connectivity:
- `mbpt::lst` now uses explicit commutators, [A,B] = AB - BA by default. Can be switched to the cheaper connected form with an option in `LSTOptions`

The `CC` class is unaffected: it passes its connectivity and its lst form explicitly.

### The change:
Keep the old behavior by asking for it:
```cpp
auto hbar = lst(H(), T(2), 4, {.use_connected_form = true});
auto eqs = ref_av(P(2) * hbar, {.connect = default_op_connections()});
```
**Compatibility**: any out-of-tree caller relying on `lst()`'s previous default (no `LSTOptions` passed) moves from the connected-product form to explicit commutators. See https://github.com/ValeevGroup/mpqc4/pull/795. 

**Note**: initially I had planned to change the default behavior of `ref_av`/`vac_av` calls, but turned out to be bad idea. So only `lst` is changed and git history is cleaned. 

Documentation on `mbpt::lst` is moved out of the `CC` class file and `LSTOptions` are now documented.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -367,6 +367,7 @@ set(SeQuant_symb_src
         SeQuant/core/tensor_network/v2.hpp
         SeQuant/core/tensor_network/v3.cpp
         SeQuant/core/tensor_network/v3.hpp
+        SeQuant/core/utility/aggregate.hpp
         SeQuant/core/utility/atomic.hpp
         SeQuant/core/utility/context.hpp
         SeQuant/core/utility/debug.hpp
--- a/SeQuant/core/utility/aggregate.hpp
+++ b/SeQuant/core/utility/aggregate.hpp
@@ -18,6 +18,11 @@ class designated_init_only {
  public:
   constexpr designated_init_only(const designated_init_only&) noexcept =
       default;
+  /// declaring the copy constructor deprecates the *implicit* copy assignment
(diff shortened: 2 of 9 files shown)


evaleev:
One more thing worth adding to the compatibility note in the description:

`LSTOptions{true, false}` (positional aggregate init, no designators) compiled before and compiles now with **inverted** meaning — the second field kept its position but flipped polarity:

```cpp
lst(A, B, r, {true, false});  // before: unitary + use_commutators=false -> connected product
                              // after:  unitary + use_connected_form=false -> explicit commutator
```

Designated-initializer callers get a compile error on `.use_commutators`, which is the right kind of break. Positional ones change behavior silently. Nothing in-tree does this and it's unlikely downstream, but since the description already carries a migration note for the default change, this is worth a line next to it.

ajay-mk:
> LSTOptions{true, false} (positional aggregate init, no designators) compiled before and compiles now with inverted meaning

I did not notice that. We can disable the positional aggregate initialization with a trick: https://godbolt.org/z/YhT4fKcKo, and only support designated/named initialization.

evaleev:
Sounds good to me

evaleev (approved):

evaleev (approved):

PR #585, opened 2026-08-04
Require designated initializers for all Options-style aggregates

evaleev:
Stacked on #584 — **base is `ajay/flip-lst-defaults`**, since `SEQUANT_DESIGNATED_INIT_ONLY` is introduced there. Retarget to `master` once #584 merges; review only the top commit.

## Why

#584 flipped `LSTOptions`' second field from `use_commutators` to `use_connected_form` — same position, inverted polarity. Designated-initializer callers got a compile error; positional callers (`LSTOptions{true, false}`) silently changed meaning. That prompted the guard in #584.

Every other options aggregate in SeQuant has the same failure mode. This applies the guard to all 20 of them, so the "reorder a field, silently reinterpret a caller" hazard is gone project-wide rather than fixed once for the struct that happened to hit it.

## What

`CanonicalizeOptions`, `Context::Options`, `BinarizationOptions`, `TransformSumExprOptions`, `De/SerializationOptions`, both `CSEOptions`, `OptimizeOptions`, `RooflineParams`, `CostParams`, `BatchPolicy`, `BlissGraphOptions`, `CreateGraphOptions` (v2, v3), `mbpt::Context::Options`, `CC::Options`, `OpParams`, `EVOptions`, `ClosedShellCCSpintraceOptions`.

**Excluded, deliberately:** `SimplifyOptions` has a user-provided constructor, so it is not an aggregate and designated init never applied; `OpInfo` and `Context::Defaults` are not options bundles; `external/bliss` is vendored.

## One addition to the guard

The tag gets `==` and `<=>` so a guarded aggregate can still default its own comparison operators.

(A second defect — the tag's user-declared copy constructor deprecating the *implicit* copy assignment, an error under `-Werror -Wdeprecated-copy` — was found while building this PR but fixed in #584 instead, since it is reachable from `LSTOptions` alone.)

## Call sites the guard rejected

- `optimize.cpp`: `CostParams` was built from `OptimizeOptions` with **seven positional fields** — correct only while two independently-editable structs keep the same field order. This is the hazard, in library code.
- `test_optimize.cpp`: 19 sites. One had already grown a `/*peak_flops_tolerance=*/` comment, which designated initializers now make real.

## Compatibility

Positional initialization of any of these types stops compiling — deliberately, and it is a hard error with a clear message, never a silent behavior change. Downstream (mpqc4) may need the same mechanical conversion.

## Verification

Clean build and **226/226 ctest** with AppleClang 17 at C++20. Guard behavior verified on clang and g++-13/15 under C++20 and C++23: `{a, b}` and `{{}, a, b}` rejected; `{}`, `{.field = a}`, copy, copy-assign, defaulted parameters, and constant expressions all fine.

### DIFF
--- a/SeQuant/core/batch_policy.hpp
+++ b/SeQuant/core/batch_policy.hpp
@@ -1,6 +1,8 @@
 #ifndef SEQUANT_CORE_BATCH_POLICY_HPP
 #define SEQUANT_CORE_BATCH_POLICY_HPP
 
+#include <SeQuant/core/utility/aggregate.hpp>
+
 #include <cstddef>
 #include <functional>
 
@@ -12,6 +14,7 @@ class Tensor;
 /// One batchability policy shared by the single-term optimizer and the runtime
 /// batched evaluator (make_evaluator, Task A3). All predicates default empty.
 struct BatchPolicy {
+  SEQUANT_DESIGNATED_INIT_ONLY;
   std::function<bool(Index const&)> is_batchable_index = {};
   /// Per-index per-batch slice size (in elements) for a batchable index -- an
   /// UPPER BOUND, not a goal. Both the single-term optimizer and the runtime
--- a/SeQuant/core/context.hpp
+++ b/SeQuant/core/context.hpp
@@ -4,6 +4,7 @@
 #include <SeQuant/core/attr.hpp>
 #include <SeQuant/core/index_space_registry.hpp>
 #include <SeQuant/core/options.hpp>
+#include <SeQuant/core/utility/aggregate.hpp>
 #include <SeQuant/core/utility/context.hpp>
 
 namespace sequant {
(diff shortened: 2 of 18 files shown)


SeQuant/core/batch_policy.hpp
```diff
@@ -1,6 +1,7 @@
 #ifndef SEQUANT_CORE_BATCH_POLICY_HPP
 #define SEQUANT_CORE_BATCH_POLICY_HPP
 
+#include <SeQuant/core/utility/aggregate.hpp>
```
ajay-mk, line 4:
```suggestion
#include <SeQuant/core/utility/aggregate.hpp>

```
Separating include groups

evaleev: Applied, thanks. I audited the other 14 headers the same script touched — `batch_policy.hpp` was the only one affected: it had no pre-existing `<SeQuant/...>` include, so my insertion fell back to the top of the std group instead of landing in a SeQuant group. Everywhere else the include went into an existing SeQuant block in alphabetical order, with the blank line before the std/third-party group already intact.

PR #591, opened 2026-08-14
Fold BraKetSymmetry::Conjugate bra<->ket orientations onto one canonical spelling and eval cache slot

kshitij-05:
A `BraKetSymmetry::Conjugate` tensor satisfies `T{p;q} = conj(T{q;p})`, but the two spellings were treated as distinct: equivalent expressions canonicalized apart, and the evaluator computed and cached both orientations.

This PR makes canonicalization pick one orientation per tensor. `Tensor` gains a conjugation flag (`conjugate()`/`conjugated()`, rendered `T^*` = the elementwise complex conjugate of `T`); when canonicalization spells a Conjugate tensor in the swapped orientation it sets the flag, so the represented value never changes. The flag participates in hashing/equality, round-trips through the serializer (`T^*{...}`), and is preserved by the mbpt transforms (csv/df/thc rules, `expand_antisymm`, `swap_spin`).

- The fold is per-tensor and always on; it runs in regular canonicalization and in TN canonicalization, both via one content-based orientation rule (`prefer_swapped_braket`), so the canonical spelling is independent of input order. (The earlier commits added an opt-in network-level bool; per the review that design was replaced, and the opt-in channel is gone — this description is of the final state.)
- Not folded: operator-valued tensors (a `NormalOperator` swap exchanges creators/annihilators), the reserved (anti)symmetrizers, and identity swaps (`T{p,q;p,q}`).
- Eval layer: a leaf hashes as its unflagged spelling, so both orientations share one cache slot; `binarize` serves a flagged leaf through `EvalOp::Adjoint`. `canonicalize_slots` reports the swap in `metadata.conj`, consumed only by the single-tensor leaf constructor.

Also in this PR:

- Core bugfix (latent, exposed here): TNV3's lexicographic dummy rewrite skipped named edges by position, so a named non-edge (pure proto) index misaligned the cutoff and a slot index got duplicated (`Edge::add_vertex` throw). Now skipped per edge; regression test included.
- `fold_conjugate_pairs_of_real_sum`: folds `{s, adjoint(s)}` summand pairs of a real-valued sum into `2*s` (used downstream by MPQC's Kramers-restricted CC).
- The ToT leaf constructor passes `default_idxptr_slottype_lesscompare{}` explicitly instead of `{}` (an empty `std::function` silently selected the space-only fallback comparator).

Behavior changes to note: since the fold is unconditional, non-Hermitian tensors (e.g. amplitudes) must be declared `Nonsymm` — the test suite was swept accordingly; `Tensor::canonicalize()` now canonicalizes even when no canonicalizer was registered (`DefaultTensorCanonicalizer` seeded under `L""`); the shared TN test runs TNV3-only (V1/V2 cannot ingest folded spellings, hidden as `[.legacy-tn]`).


SeQuant/core/tensor_network/v3.hpp
```diff
@@ -282,6 +282,16 @@ class TensorNetworkV3 {
...
+
+    /// antilinear byproduct of canonicalization: true if reaching canonical
+    /// form required complex-conjugating the network relative to its input.
+    /// Only ever set when canonicalization is asked to exploit
+    /// BraKetSymmetry::Conjugate (see canonicalize_slots' exploit_conjugate
+    /// parameter): a Hermitian (Conjugate) tensor satisfies
+    /// T{bra;ket} = conj(T{ket;bra}), so folding its two orientations onto one
+    /// canonical form carries a conjugation, recorded here (cf. `phase`, which
+    /// carries the ±1 linear byproduct of antisymmetric slot reorderings).
+    bool conj = false;
```
Krzmbrzl, line 295: I feel like a simple boolean flag is insufficient as it might only be a single tensor that needs conjugation whereas others are left unchanged.

SeQuant/core/tensor_network/v3.cpp
Krzmbrzl, line 1: It seems like support for conjugate canonicalization is only added to `canonicalize_slots` which seems to imply that regular canonicalization still doesn't handle it. If this is true, I think this is inconsistent and should be changed to also support it in regular canonicalization routines.

SeQuant/core/tensor_network/v3.cpp
```diff
@@ -889,6 +891,49 @@ TensorNetworkV3::canonicalize_slots(
...
+  // Conjugate tensor were colored identically (create_graph above), so bliss
+  // may have canonicalized it in the bra<->ket-swapped orientation. Since
+  // T{bra;ket} = conj(T{ket;bra}) for a Conjugate tensor, that swap contributes
+  // a complex conjugation. Detect it by comparing the canonical positions of
+  // each Conjugate tensor's bra- and ket-bundle vertices -- exactly the
+  // comparison canonicalize() uses for its explicit swap
+  // (canonical_bra_ket_bundle_order, v3.cpp above). N.B. metadata.conj is a
+  // single network-level bit: rigorous for one Conjugate tensor (the
+  // proto-indexed-leaf case that eval-node identities canonicalize), the
+  // per-leaf conjugation of a multi-Conjugate-tensor network is future work.
```
Krzmbrzl: is that actually much more complicated? If not, it would make sense to just do the full thing here 🤔

SeQuant/core/tensor_network/v3.hpp
```diff
@@ -297,12 +307,19 @@ class TensorNetworkV3 {
...
+  /// distinctly colored (historical behavior), so it is a no-op unless
+  /// requested.
   /// @return the computed canonicalization metadata
   SlotCanonicalizationMetadata canonicalize_slots(
       const container::vector<std::wstring> &cardinal_tensor_labels = {},
       const NamedIndexSet *named_indices = nullptr,
       SlotCanonicalizationMetadata::named_index_compare_t named_index_compare =
-          default_idxptr_slottype_lesscompare{});
+          default_idxptr_slottype_lesscompare{},
+      bool exploit_conjugate = false);
```
Krzmbrzl: I think it's time to bundle these options into an options struct and pass that instead. The number of parameters starts becoming unwieldy 👀

SeQuant/core/tensor_canonicalizer.cpp
```diff
@@ -332,39 +332,56 @@ template <typename Callable, typename... Args>
...
   // spaces (e.g. g{p,q;r,s}) compare equal and are left untouched; only
   // differing-color bundles are reoriented (so e.g. a half-tensor X{;a;x} folds
   // into X{a;;x}). Mirrors the bra<->ket bundle swap in
   // TensorNetworkV3::canonicalize_slots.
+  const TensorBlockIndexComparer cmp;
+  auto space_less = [&cmp](const Index& a, const Index& b) {
+    return cmp.compare_spaces(a, b) < 0;
+  };
+  auto bra = mutable_bra_range(t);
+  auto ket = mutable_ket_range(t);
```
Krzmbrzl: Why does this need `mutable_*`? It seems like they aren't actually mutated

SeQuant/core/tensor_canonicalizer.cpp
```diff
@@ -332,39 +332,56 @@ template <typename Callable, typename... Args>
...
+  // Compare the bundles by their space sequences *sorted by color*, so the
+  // decision is independent of the within-bundle index order. Column/perm
+  // symmetry can permute the bra (and ket) order without changing the tensor,
+  // and a comparison over the as-given order could otherwise pick different
+  // orientations for equivalent inputs.
+  std::vector<Index> bra_spaces, ket_spaces;
+  for (auto&& idx : bra) bra_spaces.push_back(idx);
+  for (auto&& idx : ket) ket_spaces.push_back(idx);
```
Krzmbrzl: Using universal reference is unnecessary if we don't `forward` the indices

```suggestion
  for (const auto& idx : bra) bra_spaces.push_back(idx);
  for (const auto& idx : ket) ket_spaces.push_back(idx);
```

tests/unit/test_tensor_network.cpp
```diff
@@ -243,6 +243,50 @@ TEMPLATE_TEST_CASE("tensor_network_shared", "[elements]", TensorNetworkV1,
...
+    SECTION("exploit_conjugate") {
+      if constexpr (TN::version() >= 3) {
+        // A Hermitian (BraKetSymmetry::Conjugate) tensor satisfies
+        //   h{bra;ket} = conj(h{ket;bra}),
+        // so its two bra<->ket orientations should fold onto a single canonical
+        // form carrying a recorded conjugation -- but only when the caller opts
+        // in via exploit_conjugate. Default keeps Conjugate bra/ket distinctly
+        // colored (no fold, no conj), mirroring historical behavior.
+        const auto cardinal = TensorCanonicalizer::cardinal_tensor_labels();
+        auto md = [&cardinal](const std::wstring& s, bool exploit) {
```
Krzmbrzl: `md` seems like a very non-descriptive name. It's short but also doesn't hint at what the function is doing. I'd recommend a more readable/speaking function name

Krzmbrzl (commented):
I feel like exploitation of conjugate BraKet symmetry should be on-by-default. That would be consistent with how we handle all other symmetries.

Also, I don't think tracking the conjugation result by means of a boolean variable is the right way to go. At some point in the future (ideally already within this PR) we want to handle this to full extent implying that only some tensors might be conjugated whereas others might not. Hence, we need a more granular result telling us exactly which tensors need conjugation. Otherwise, we will have to break the public API eventually to account for this.

I have implemented a `TreeIndex` class on a different branch that allows uniquely identifying an element in a (potentially nested) (expression) tree. We could use that and then return a vector of `TreeIndex` objects to specify which tensors need conjugation. This would have the advantage that this interface would still be usable once the canonicalizer can deal with nested expressions.

SeQuant/core/tensor_network/v3.cpp
```diff
@@ -889,6 +891,49 @@ TensorNetworkV3::canonicalize_slots(
...
+          const bool bra = vt == VertexType::TensorBraBundle;
+          bundle_pos[tensor_ord][bra ? 0 : 1] = canonize_perm[v];
+        }
+      }
+    }
+    bool conj = false;
+    for (const auto &[tensor_ord, bk] : bundle_pos) {
+      // bra bundle canonically after ket bundle => canonical form is the
+      // bra<->ket-swapped (conjugated) orientation of the input.
+      if (bk[0] && bk[1] && *bk[0] > *bk[1]) conj = !conj;
```
ajay-mk: I see the comment which says "multi-Conjugate-tensor network is future work". The current version will break if two tensors are conjugated in a network. Agree with Robert that one single bool cannot represent this.

SeQuant/core/eval/eval_expr.cpp
```diff
@@ -426,6 +465,48 @@ EvalExprNode binarize(Tensor const& t) {
     return EvalExprNode{std::move(adj), std::move(bare_leaf),
                         std::move(sentinel)};
   }
+
+  // Opt-in: fold the two bra<->ket orientations of a BraKetSymmetry::Conjugate
```
ajay-mk: The code block above has very similar structure to this. Maybe factor out the common piece.

SeQuant/core/eval/eval_expr.hpp
```diff
@@ -74,7 +74,17 @@ class EvalExpr {
   ///
   /// \brief Construct an EvalExpr object from a tensor.
   ///
-  explicit EvalExpr(Tensor const& tnsr);
+  /// \param tnsr The tensor to wrap as a leaf.
+  /// \param exploit_conjugate If true, a proto-indexed (ToT) leaf is
```
ajay-mk: I assume this line is wrong? The logic is supported for non-ToT case also right?

tests/unit/test_eval_expr.cpp
```diff
@@ -557,3 +558,129 @@ TEST_CASE("eval_expr", "[EvalExpr]") {
     REQUIRE_NOTHROW(result_expr(t1, t2, EvalOp::Product));
   }
 }
+
+// B2: the opt-in exploit_conjugate path folds the two bra<->ket orientations of
```
ajay-mk: "B2" can be removed

ajay-mk (commented):
I think this can be a default-on feature. A few things that concerns me:

- Since this affects the export pipeline (if opted-in), we should include some tests and check if the produced code makes sense. 
- Minor: I think the "flat tensor/array" language for proto-index free tensors is conflicting because we refer to them as Tensor of Scalars also. We should stick to one, maybe out of scope of this PR.

SeQuant/core/tensor_network/v3.hpp
kshitij-05: Yes, a boolean-only flag was not a good idea, and I realized this while working on another draft PR where I had multiple conjugated tensors in the same TN. Now the tensor label itself carries the conj flag.

SeQuant/core/tensor_network/v3.cpp
kshitij-05: Yes, `TensorBlockCanonicalizer` is now extended, and slot-level (TN) and tensor-level canonicalization now produce consistent orientations

SeQuant/core/tensor_network/v3.cpp
kshitij-05: does the full thing now

SeQuant/core/tensor_network/v3.hpp
kshitij-05: I just removed the need to keep track of these. So we no longer need to bundle these options

SeQuant/core/tensor_canonicalizer.cpp
kshitij-05: switched to const now

SeQuant/core/tensor_network/v3.cpp
kshitij-05: I redesigned the conj handling see the comments above

tests/unit/test_eval_expr.cpp
kshitij-05: removed

SeQuant/core/eval/eval_expr.hpp
kshitij-05: Yeah, it was. Thanks for pointing it out. Rewritten

SeQuant/core/eval/eval_expr.cpp
kshitij-05: deduped. Thank you

SeQuant/core/tensor_network/v3.cpp
```diff
@@ -366,65 +366,86 @@ ExprPtr TensorNetworkV3::canonicalize_graph(const NamedIndexSet &named_indices,
...
+  // canonical bra<->ket orientation of braket-foldable tensors
   for (std::size_t i = 0; i < tensors_.size(); ++i) {
     AbstractTensor &tensor = *tensors_[i];
 
-    if (column_symmetry(tensor) != ColumnSymmetry::Symm) continue;
+    const bool column_symm = column_symmetry(tensor) == ColumnSymmetry::Symm;
     const auto asymm = symmetry(tensor) == Symmetry::Nonsymm;
 
-    if (asymm) {  // asymmetric tensor? order column slots only
+    if (column_symm && asymm) {  // asymmetric tensor? order column slots only
```
Krzmbrzl, line 378: The comment no longer matches the code as this branch now also deals with column symmetric tensors

SeQuant/core/tensor_network/v3.cpp
```diff
@@ -366,65 +366,86 @@ ExprPtr TensorNetworkV3::canonicalize_graph(const NamedIndexSet &named_indices,
...
+    // bundle vertices are ordered by bliss tie-breaking, which depends on the
+    // input presentation. Identical bundles (diagonal trace T{p,q;p,q})
+    // never swap, so no spurious conjugation marker can arise on an identity
+    // swap.
+    if (prefer_swapped_braket(tensor)) {
       tensor._swap_bra_ket();
+      // for a Conjugate tensor the swapped spelling denotes the conjugate
+      // value (T{q;p} = conj(T{p;q})): keep the represented value invariant
+      // by toggling the elementwise-conjugation marker
+      if (bksymm == BraKetSymmetry::Conjugate) tensor._conjugate();
```
Krzmbrzl, line 448: What's the reason for not using the braket ordering dictated by the canonicalized graph? There were quite a few subtleties involved in which parts of the graph can be ignored so unless there is a real good reason not to, I'd strongly vote to keep swapping bra and ket based on graph properties.

SeQuant/core/tensor_network/v3.cpp
```diff
@@ -601,10 +622,15 @@ ExprPtr TensorNetworkV3::canonicalize(
...
 
     // Use the new order of edges as the canonical order of indices and relabel
-    // accordingly (but only anonymous indices, of course)
-    for (std::size_t i = named_indices.size(); i < edges_.size(); ++i) {
+    // accordingly (but only anonymous indices, of course). Skip named indices
+    // by CHECKING each edge, not by starting the loop at named_indices.size():
+    // a named index that is not an edge (e.g. a pure proto index) would shift
+    // that positional cutoff onto an anonymous edge, whose skipped ordinal the
+    // factory would then hand to another same-space edge -- a non-injective
+    // rewrite that duplicates a slot index.
```
Krzmbrzl, line 630: This comment is not useful - it's an explanation for the change and thus useful in the associated commit message but it doesn't give someone reading the code any real information given that they don't see a diff but only the new code. Hence, the comment reasons about things that are no longer there 👀 

Instead, we should just have a test that makes sure that pure proto indices canonicalize as expected (aka: that fails with the old code and works with the new code). That will make sure nobody changes it back without noticing the issue.

SeQuant/core/tensor_network/v3.hpp
Krzmbrzl: The flag is still there though - what it is used for now?

SeQuant/core/tensor_canonicalizer.cpp
```diff
@@ -317,9 +326,74 @@ void DefaultTensorCanonicalizer::tag_indices(AbstractTensor& t) const {
   });
 }
 
+bool braket_orientation_pinned(const AbstractTensor& t) {
+  const auto lbl = t._label();
+  return lbl == reserved::antisymm_label() || lbl == reserved::symm_label() ||
+         lbl == reserved::transposition_label();
+}
```
Krzmbrzl, line 333: I think we can/should include the cnumber check into this function. That seems like it could avoid future errors as calling sites would only have to remember to call this function rather than to _also_ check whether a tensor is a cnumber or not.

PR #595, opened 2026-08-22
LTO support

Krzmbrzl:


cmake/compiler.cmake
```diff
@@ -60,3 +60,66 @@ function(target_set_warning_flags TARGET)
...
+	elseif(TARGET_TYPE STREQUAL "STATIC_LIBRARY" OR TARGET_TYPE STREQUAL "OBJECT_LIBRARY")
+		# For static/object libraries we only want to enable LTO by default, if we can create
+		# "fat" object files. Those can still be linked without LTO and hence shouldn't
+		# break any downstream use.
+		set(ENABLE_LTO ${LTO_FLAG_SUPPORTED})
+	elseif(LTO_FLAG_SUPPORTED OR CMAKE_SUPPORTS_COMPILER_LTO)
+		# Anything but static/object libraries is also linked by us and
+		# hence enabling LTO doesn't affect downstream compatibility
+		set(ENABLE_LTO ON)
+	endif()
```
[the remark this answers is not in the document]
Krzmbrzl: This is a semicolon-separated list and explicitly supported and recommended by the docs. Plus, it works as-is

ajay-mk:
@Krzmbrzl Could you update the CMake options table in docs?

tests/unit/test_optimize.cpp
```diff
@@ -727,7 +727,7 @@ TEST_CASE("optimize", "[optimize]") {
       container::svector<Index> targets;
       auto aux = opt::detail::batchable_index_list(net, is_batchable);
       std::size_t const m = aux.size();
-      auto batch_fn = [batch](Index const&) -> std::size_t { return batch; };
+      auto batch_fn = [](Index const&) -> std::size_t { return batch; };
```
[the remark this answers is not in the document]
Krzmbrzl: Well, it _does_ compile and in fact it was a compiler warning that caused me to remove the (according to the compiler) unnecessary capture.

evaleev (commented):
## Code review: LTO support

Reviewed the full diff (CI action bumps, the new `target_set_optimization_flags`, benchmark `OpRegistry` fixes + "quick" mode, dependency bumps, warning fixes). 8 findings inline, ordered by severity below; the LTO flag-detection claims were verified empirically on Apple Clang 17 with a standalone CMake reproduction.

| # | Where | Severity |
|---|---|---|
| 1 | `cmake/compiler.cmake:95` | medium/high |
| 2 | missed `target_set_optimization_flags` consumers | medium |
| 3 | `cmake/compiler.cmake:67` (+ `benchmarks/CMakeLists.txt:23`) | medium |
| 4 | `benchmarks/coupled_cluster.cpp:28` (`maxRank`) | medium |
| 5 | `benchmarks/wick.cpp:131` (`nMbptInputs`) | low/medium |
| 6 | `cmake/compiler.cmake:122` | low |
| 7 | `cmake/compiler.cmake:81,84-86` | low |
| 8 | `benchmarks/coupled_cluster.cpp:18` | low |

### Verified as correct — no action needed
- `SeQuant/core/eval/eval.hpp:560` — `ResultPtr left/right = {}` is safe: `Frame` remains an aggregate and all three construction sites (581, 656, 695) use designated initializers.
- `tests/unit/test_optimize.cpp:730,772,814` — dropping the `batch` capture compiles; all three are `std::size_t const batch = 1;`, so the read is a constant expression and not an odr-use.
- `--benchmark_dry_run=true` exists in the tracked Google Benchmark (`src/benchmark.cc:764` in v1.9.4, and `SEQUANT_OLDEST_GOOGLEBENCHMARK_VERSION` is 1.9.3), and `build/benchmarks/sequant_benchmarks` is the correct output path (no `CMAKE_RUNTIME_OUTPUT_DIRECTORY` is set).
- `check_cxx_compiler_flag("-flto;-ffat-lto-objects" ...)` does pass *both* flags — `Internal/CheckCompilerFlag.cmake` expands the list unquoted into `CMAKE_REQUIRED_DEFINITIONS`.
- `make_minimal_registry()` covers everything the fixed benchmarks need (`t`, `h`, `f`, `g`, plus `L`/`R` backing `t::l`/`t::r`), and `spintrace` genuinely needs no registry — so `benchmarks/spintrace.cpp` correctly went untouched.
- `SEQUANT_QUICK_BENCHMARKS` being `option()`-declared inside `benchmarks/` is fine with the CI's unconditional `-D...=ON`; it only produces an "unused variable" notice in the valgrind/sanitize jobs where benchmarks are off.

### Not verified
The `external/versions.cmake` tag bumps (utfcpp v4.1.1, CLI11 v2.7.2, spdlog v1.17.0, Catch2 v3.15.3, googlebenchmark v1.9.5, pybind11 v3.1.0) could **not** be checked — this review ran without network access, so `git ls-remote` against the upstream repos was unavailable. Someone should confirm those tags exist and that the `SEQUANT_OLDEST_*` floors are still accurate. Same for the `actions/*@v7` / `@v6` / `@v9` bumps in the workflows.

cmake/compiler.cmake
```diff
@@ -60,3 +60,66 @@ function(target_set_warning_flags TARGET)
...
+	check_cxx_compiler_flag("-flto;-ffat-lto-objects" FAT_LTO_FLAG_SUPPORTED)
+
+	if (DEFINED SEQUANT_LTO)
+		# Always honor explicit user choice
+		set(ENABLE_LTO ${SEQUANT_LTO})
+	elseif(TARGET_TYPE STREQUAL "STATIC_LIBRARY" OR TARGET_TYPE STREQUAL "OBJECT_LIBRARY")
+		# For static/object libraries we only want to enable LTO by default, if we can create
+		# "fat" object files. Those can still be linked without LTO and hence shouldn't
+		# break any downstream use.
+		set(ENABLE_LTO ${LTO_FLAG_SUPPORTED})
```
evaleev: **[1 — medium/high] The static/object-library gate uses the wrong variable, defeating its own stated purpose.**

The comment right above says LTO is enabled here *only* if we can create fat object files, so the archives "can still be linked without LTO and hence shouldn't break any downstream use". But this sets `ENABLE_LTO` from `LTO_FLAG_SUPPORTED`, not `FAT_LTO_FLAG_SUPPORTED`. `FAT_LTO_FLAG_SUPPORTED` is only ever consulted at line 118 to *add* `-ffat-lto-objects` — it never gates anything.

Verified on Apple Clang 17 with a standalone repro:
```
-- Performing Test FAT_LTO_FLAG_SUPPORTED - Failed
-- IPO=YES LTO=1 AUTO=1 FAT=

$ clang++ -c -flto -ffat-lto-objects t.cpp
clang++: error: unsupported option '-ffat-lto-objects' for target 'arm64-apple-darwin25.3.0'
$ clang++ -c -flto t.cpp -o t2.o && file t2.o
t2.o: LLVM bitcode, wrapper
```

Scenario: any Clang/AppleClang Release build. Every `SeQuant-*` static lib (via `sequant_add_library`, `CMakeLists.txt:98`) and `SeQuant-bliss` (`CMakeLists.txt:276`) is archived from bitcode-only objects with no fat fallback. These are installed and exported as `SeQuant::symb`, `SeQuant::bliss`, etc.; a downstream consumer linking them without an LTO-capable linker/plugin gets `file format not recognized` / `archive has no index` / undefined symbols — exactly the breakage this branch was written to prevent.

Fix: `set(ENABLE_LTO ${FAT_LTO_FLAG_SUPPORTED})`.

tests/integration/CMakeLists.txt
```diff
@@ -56,6 +56,7 @@ foreach(current IN LISTS TEST_CASES)
     endif()
 
     target_set_warning_flags("${test_name}")
+    target_set_optimization_flags("${test_name}")
```
evaleev, line 59: **[2 — medium] Several consumers of the newly-LTO'd static libs never get `target_set_optimization_flags`.** (General finding; anchored here because this is one of the places the call *was* correctly added.)

The PR adds the call to `utilities/`, `tests/unit/`, this file, and `benchmarks/`, but these link `SeQuant` static libraries and were missed, so they link `-flto`-compiled archives with no `-flto` at link time:

- `python/CMakeLists.txt:15` — `python-sequant` (a `MODULE` library linking `SeQuant`)
- `doc/examples/CMakeLists.txt:26`
- `tests/integration/eval/CMakeLists.txt:19,38,57` — `eval_ta`, `eval_btas`, `eval_tapp`

Combined with finding 1 (bitcode-only archives on Clang), this is the in-tree failure path: Linux + Clang + `-DCMAKE_BUILD_TYPE=Release -DSEQUANT_PYTHON=ON`, where GNU `ar`/`ld` without the LLVMgold plugin cannot index or read those archives. CI dodges it only by accident — Ubuntu uses g++-14 (which *does* support `-ffat-lto-objects`) and macOS uses ld64 (which reads bitcode natively).

Fix: add `target_set_optimization_flags(...)` to those targets too (and/or fix finding 1, which removes the hazard at the source).

cmake/compiler.cmake
```diff
@@ -60,3 +60,66 @@ function(target_set_warning_flags TARGET)
         target_compile_options("${TARGET}" PRIVATE "-Wno-unused-lambda-capture")
     endif()
 endfunction()
+
+
+
+function(target_set_optimization_flags TARGET)
+	if (CMAKE_BUILD_TYPE STREQUAL "Debug")
```
evaleev, line 125: **[3 — medium] The `CMAKE_BUILD_TYPE STREQUAL "Debug"` guard is a no-op under multi-config generators.**

With Ninja Multi-Config, Xcode, or Visual Studio, `CMAKE_BUILD_TYPE` is empty, so this early return never fires and `-flto=auto` is applied unconditionally to *every* config, including Debug — slowing Debug builds substantially and degrading debuggability.

The identical guard at `benchmarks/CMakeLists.txt:23` has the same problem in reverse: under a multi-config generator, `SEQUANT_BENCH_MAX_CC_RANK` / `SEQUANT_BENCH_MAX_WICK_INPUTS` are never defined for the Debug config, so the "don't let Debug benchmarks take forever" protection silently disappears.

Fix: use a `$<CONFIG:Debug>` generator expression for the flags (e.g. wrap the options in `$<$<NOT:$<CONFIG:Debug>>:...>`) rather than testing `CMAKE_BUILD_TYPE` at configure time.

benchmarks/coupled_cluster.cpp
```diff
@@ -1,15 +1,22 @@
 #include <benchmark/benchmark.h>
 
+#include <SeQuant/domain/mbpt/context.hpp>
 #include <SeQuant/domain/mbpt/models/cc.hpp>
 
+#ifdef SEQUANT_BENCH_MAX_CC_RANK
+static constexpr std::size_t maxRank = SEQUANT_BENCH_MAX_CC_RANK;
+#else
 static constexpr std::size_t maxRank = 10;
```
evaleev, line 8: **[4 — medium] Now that `cc_full_derivation` actually runs, the un-capped default `maxRank = 10` makes the `benchmark_compare` workflow effectively non-terminating.** (Applies to this `maxRank = 10` and its use at line 28, `BENCHMARK(cc_full_derivation)->DenseRange(2, maxRank)`.)

Before this PR the benchmark threw immediately (missing `OpRegistry`), so `DenseRange(2, 10)` was never actually exercised. The registry fix at line 18 makes it run for real.

`bin/admin/benchmark_compare.py:189-195` configures with `-DSEQUANT_BENCHMARKS=ON` but **not** `-DSEQUANT_QUICK_BENCHMARKS=ON`, and `run_benchmarks` invokes the binary with no `--benchmark_filter`. So `benchmark_compare.yml` will now attempt a full CC rank-10 t-amplitude derivation. Extrapolating from this repo's own documented timing (`doc/user/getting_started/operators.rst:89`: rank 4 -> 0.107 s, with roughly order-of-magnitude growth per rank), rank 8-10 is hours-to-weeks. The job will hang or time out on every PR.

Fix: either cap `maxRank` unconditionally at something feasible (~5-6), or have `benchmark_compare.py` add `-DSEQUANT_QUICK_BENCHMARKS=ON` to `cmake_variables`.

benchmarks/wick.cpp
```diff
@@ -13,7 +13,11 @@
 using namespace sequant;
 using namespace sequant::mbpt;
 
+#ifdef SEQUANT_BENCH_MAX_WICK_INPUTS
+static constexpr std::size_t nInputs = SEQUANT_BENCH_MAX_WICK_INPUTS;
```
evaleev, line 17: **[5 — low/medium] `SEQUANT_BENCH_MAX_WICK_INPUTS` caps `nInputs` but not `nMbptInputs`, leaving the most expensive benchmarks in the file uncapped.**

`static constexpr std::size_t nMbptInputs = 4;` (line 131) is unconditional, and the `mbpt_vac_av` family (lines 176-178: two variants x 4 inputs) is the heaviest set here — case 3 is `A(-5) * h(2) * t(2) * t(2) * t(3)`.

Scenario: the new `Ensure benchmarks not broken` step in `cmake.yml:139` has condition `!valgrind && !sanitize`, which matches the **Debug** non-sanitize matrix entries, so these run unoptimized (and unity-built). The job's 60-minute timeout already has to cover a full build, so this is a plausible timeout source — and it contradicts the stated intent of the guard in `benchmarks/CMakeLists.txt` ("exclude benchmarks that are likely to take forever in this setup").

Fix: gate `nMbptInputs` on the same macro (or add a `SEQUANT_BENCH_MAX_MBPT_INPUTS`).

cmake/compiler.cmake
```diff
@@ -60,3 +60,66 @@ function(target_set_warning_flags TARGET)
...
+			else()
+				target_compile_options("${TARGET}" PRIVATE -flto)
+				target_link_options("${TARGET}" PRIVATE -flto)
+			endif()
+
+			if (FAT_LTO_FLAG_SUPPORTED)
+				target_compile_options("${TARGET}" PRIVATE -ffat-lto-objects)
+			endif()
+		else()
+			set_target_properties("${TARGET}" PROPERTIES INTERPROCEDURAL_OPTIMIZATION ON)
```
evaleev, line 194: **[6 — low] Forced `SEQUANT_LTO=ON` on a toolchain with no LTO support is a hard configure error.**

If the user passes `-DSEQUANT_LTO=ON` while both `LTO_FLAG_SUPPORTED` and `CMAKE_SUPPORTS_COMPILER_LTO` are false, line 90 sets `ENABLE_LTO=ON`, the `if (LTO_FLAG_SUPPORTED)` at 103 is false, and this branch sets `INTERPROCEDURAL_OPTIMIZATION ON` on a compiler CMake already reported as unable to do IPO. CMake then aborts at generate time with `INTERPROCEDURAL_OPTIMIZATION is not supported`.

Fix: intersect the explicit-user-choice branch with actual capability, or emit an explicit `message(FATAL_ERROR ...)` that names `SEQUANT_LTO` so the cause is obvious.

cmake/compiler.cmake
```diff
@@ -60,3 +60,66 @@ function(target_set_warning_flags TARGET)
...
+
+	if (TARGET_TYPE STREQUAL "INTERFACE_LIBRARY")
+		message(WARNING "target_set_optimization_flags is not intended to be used on interface targets")
+		return()
+	endif()
+
+	include(CheckCXXCompilerFlag)
+	include(CheckIPOSupported)
+
+	check_ipo_supported(RESULT CMAKE_SUPPORTS_COMPILER_LTO LANGUAGES CXX)
```
evaleev: **[7 — low] Result variables leak into the global cache under generic / reserved names.**

`check_ipo_supported` writes to `CMAKE_SUPPORTS_COMPILER_LTO`, which squats on the `CMAKE_`-reserved prefix, and lines 84-86 cache `LTO_FLAG_SUPPORTED`, `LTO_AUTO_SUPPORTED`, `FAT_LTO_FLAG_SUPPORTED` under unprefixed names.

Scenario: SeQuant consumed via `add_subdirectory`/`FetchContent` by a parent that already defines any of these. A *normal* (non-cache) variable in the parent shadows the cache entry inside this function, so the parent silently drives SeQuant's LTO decision — including flipping it on for static libs on a toolchain where the check would have said no.

Fix: prefix them `SEQUANT_*` (e.g. `SEQUANT_LTO_FLAG_SUPPORTED`).

benchmarks/coupled_cluster.cpp
```diff
@@ -1,15 +1,22 @@
...
+#endif
 
 using namespace sequant;
 using namespace sequant::mbpt;
 
 static void cc_full_derivation(benchmark::State &state) {
   const std::size_t rank = state.range(0);
 
+  set_default_mbpt_context({.op_registry_ptr = make_minimal_registry()});
```
evaleev: **[8 — low] Global MBPT context mutated from inside a benchmark body.**

This is the unscoped `set_default_mbpt_context`, unlike `wick.cpp:155` which correctly uses `set_scoped_default_mbpt_context`. Three consequences:

1. It permanently replaces the process-wide default context for every benchmark that runs afterwards.
2. It re-allocates a registry on every invocation of the benchmark function — once per `DenseRange` value, per repetition.
3. Any future benchmark that needs a registry will work or fail depending on cross-TU registration order relative to this one.

`benchmarks/main.cpp` already does the global setup (`set_default_context` for both statistics) and is the natural home for this — one `set_default_mbpt_context({.op_registry_ptr = make_minimal_registry()})` there covers every TU deterministically.

tests/unit/test_optimize.cpp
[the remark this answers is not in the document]
Krzmbrzl: https://stackoverflow.com/a/43468519

cmake/compiler.cmake
Krzmbrzl: Correct though I don't think the added complexity is really worth the hassle. This can be dealt with once someone actually uses Windows to do SeQuant developing.

benchmarks/coupled_cluster.cpp
Krzmbrzl: This is not related to this PR - it's working as initially intended. @ajay-mk I'll leave the decision about this one up to you.

benchmarks/wick.cpp
Krzmbrzl: Seems to cause no issue in practice

cmake/compiler.cmake
Krzmbrzl: If a user makes an explicit decision, they will have to live with the consequences of that decision.

benchmarks/coupled_cluster.cpp
Krzmbrzl: e976a19e4568a22d39e53aa82fdd456463047a55

cmake/compiler.cmake
Krzmbrzl: 73ca17fe402dfbb7c4a9549e298b21309c9a1f3b

tests/integration/CMakeLists.txt
Krzmbrzl: 803ac19a18f3c428e033b4edbdb67c3173bf84b2

cmake/compiler.cmake
Krzmbrzl: 7460c7ecd9d7a0cbf54e1be4e03f6206a26e837c

benchmarks/coupled_cluster.cpp
ajay-mk: `maxRank = 10` is fine, and it was actually exercised before, contrary to what the comment says. Since we use a Release build for that script, it is fine to keep `maxRank` and 10 and should go through within the time limit.

cmake/compiler.cmake
```diff
@@ -60,3 +60,66 @@ function(target_set_warning_flags TARGET)
...
+
+	if (DEFINED SEQUANT_LTO)
+		# Always honor explicit user choice
+		set(ENABLE_LTO ${SEQUANT_LTO})
+	elseif(TARGET_TYPE STREQUAL "STATIC_LIBRARY" OR TARGET_TYPE STREQUAL "OBJECT_LIBRARY")
+		# For static/object libraries we only want to enable LTO by default, if we can create
+		# "fat" object files. Those can still be linked without LTO and hence shouldn't
+		# break any downstream use.
+		set(ENABLE_LTO ${SEQUANT_FAT_LTO_FLAG_SUPPORTED})
+	elseif(SEQUANT_LTO_FLAG_SUPPORTED OR CMAKE_SUPPORTS_COMPILER_LTO)
```
evaleev: **[A — medium-low] LTO *link* options are enabled by a probe that never links.**

`a1f1340` added `set(CMAKE_TRY_COMPILE_TARGET_TYPE "STATIC_LIBRARY")` at line 83, which makes all three `check_cxx_compiler_flag` probes compile-only — the commit message states the assumption outright ("under the assumption that if the compiler supports the flag, we will encounter a linker that can handle LTO").

This line then enables LTO on `SEQUANT_LTO_FLAG_SUPPORTED` **OR** `CMAKE_SUPPORTS_COMPILER_LTO`, and lines 108/112 put `-flto=auto` / `-flto` into `target_link_options`. So the decision to hand LTO flags to the *linker* is made by a probe that never invoked the linker, and the one check that does link (`check_ipo_supported`, line 81) is OR'd away.

Scenario: GCC or Clang paired with a GNU `ld` that has no LTO plugin — a GCC built against plugin-less binutils, or Clang defaulting to bfd `ld` with no `LLVMgold.so`. `-flto` compiles fine, so the probe passes, and then every SeQuant executable / shared / module target fails to link in Release with `plugin needed to handle lto object` or a wall of undefined symbols. Nothing in the configure output hints at the cause.

Fix: `AND` on `CMAKE_SUPPORTS_COMPILER_LTO` here, or restore an `EXECUTABLE` try-compile target type for the plain `-flto` probe (keeping `STATIC_LIBRARY` only for the fat-objects probe, where no link is involved anyway).

cmake/compiler.cmake
evaleev: **[B — low] `check_ipo_supported` re-runs a full nested `try_compile` on every call, and it is now called 30-40 times per configure.**

`CheckIPOSupported.cmake` deliberately does not cache — it ends with `unset(_IPO_LANGUAGE_CHECK_RESULT CACHE)` — so each invocation configures *and builds* a throwaway CMake project. This PR takes the call count from 0 to 20 literal call sites plus the loop bodies in `utilities/CMakeLists.txt` and `tests/integration/CMakeLists.txt`.

Measured on an M-series mac with this branch's `compiler.cmake` verbatim, same project, only the number of `target_set_optimization_flags` calls varying:

```
 1 call  -> 1.60 s total configure
30 calls -> 14.35 s total configure     (~0.44 s per extra call)
```

That is roughly 13-18 s added to *every* `cmake` run, for a result that cannot change between calls. The `check_cxx_compiler_flag` probes below are fine — those do cache.

Fix: wrap all four checks in a one-shot guard, e.g. `if (NOT DEFINED SEQUANT_IPO_SUPPORTED) ... set(SEQUANT_IPO_SUPPORTED ${...} CACHE INTERNAL "")` , or hoist them to a single call site at include time.

python/CMakeLists.txt
```diff
@@ -13,6 +13,7 @@ endif()
 pybind11_add_module(python-sequant MODULE src/sequant/_sequant.cc)
 
 target_link_libraries(python-sequant PRIVATE SeQuant)
+target_set_optimization_flags(python-sequant)
```
evaleev: **[C — low] This line is my fault, and it should be dropped.**

I asked for it in finding 2 of the previous review without checking that pybind11 already handles LTO for this target. It does: `pybind11_add_module` does `target_link_libraries(${target_name} PRIVATE pybind11::lto)` (`pybind11NewTools.cmake:324`) unless `NO_EXTRAS`, and `pybind11::lto` carries `INTERFACE_COMPILE_OPTIONS` / `INTERFACE_LINK_OPTIONS` of `-flto[=thin|=auto]` plus `-fno-fat-lto-objects` on GCC (`pybind11Common.cmake:332-421`), already guarded by a not-Debug generator expression.

Concrete effect of adding ours on top: on GCC the compile line for `_sequant.cc` ends up with both `-ffat-lto-objects` (ours, PRIVATE) and `-fno-fat-lto-objects` (pybind11's, appended after as an interface option), plus two differently-spelled `-flto` variants. Last-one-wins makes it benign today, but it is duplicate machinery that will silently fight pybind11's the moment either side changes its spelling — and pybind11 is the side that knows what a Python extension module needs.

Recommend reverting this one line. Sorry for the churn.

doc/user/getting_started/installing.rst
```diff
@@ -110,6 +110,9 @@ Useful CMake Variables
      - ``ABORT`` in ``Debug`` mode, ``IGNORE`` otherwise
      - Controls how assertions within SeQuant's code are handled. Valid options are ``ABORT``, ``THROW`` and ``IGNORE``. The latter disables
        assertions, whereas the former keep them active and either abort the program or throw an exception on violation respectively.
+   * - SEQUANT_LTO
```
evaleev, line 113: **[D — low] The documented default does not match the code, and `SEQUANT_LTO` is not actually a cache option.**

Three things in this new row:

1. "`ON` if the compiler supportes "fat" LTO objects" is only true for STATIC/OBJECT targets (`compiler.cmake:95`). For executables, shared libraries and module libraries, `compiler.cmake:96` turns LTO on whenever plain `-flto` works, regardless of fat-object support. Directly observed in a repro on Apple clang 17: `SEQUANT_FAT_LTO_FLAG_SUPPORTED` failed, yet the executable still got `-flto=auto` on both compile and link.
2. `SEQUANT_LTO` is never declared with `option()` or `set(... CACHE ...)` anywhere in the tree — `compiler.cmake:88` only tests `if (DEFINED SEQUANT_LTO)`. So a documented user-facing switch never appears in `cmake -LH`, `ccmake`, or the GUI; a user has to know the name from this table.
3. Typo: "supportes" -> "supports".

tests/integration/eval/CMakeLists.txt
```diff
@@ -23,6 +23,7 @@ if (SEQUANT_HAS_TILEDARRAY)
     )
     set_target_properties(eval_ta PROPERTIES CXX_SCAN_FOR_MODULES OFF)
     target_link_libraries(eval_ta PRIVATE eval_shared tiledarray)
+    target_set_optimization_flags(eval_ta)
```
evaleev, line 26: **[low] Residual of prior finding 2 — anchored here because the real location has no hunks in this diff.**

Real location: `doc/examples/CMakeLists.txt:48`, which calls `target_set_warning_flags("${current_name}")` but never `target_set_optimization_flags`. It is now the only consumer of the SeQuant static libraries without it, after this file, `python/`, `utilities/`, `tests/unit/` and `tests/integration/` were all covered.

With finding 1 fixed this is no longer a breakage path — the archives are only LTO'd on toolchains where fat objects exist, so the doc-example targets link fine without `-flto`. It is purely a consistency gap now: those targets are the only ones in the tree built without the project's optimization settings.

evaleev (commented):
## Re-review at `192a150`

Status of the 8 findings from the previous review (pinned to `a1f1340`). I re-read the current source rather than the replies, and verified the LTO behaviour with a standalone CMake repro that uses this branch's `cmake/compiler.cmake` verbatim (Apple clang 17.0.0).

| # | Finding | Status |
|---|---|---|
| 1 | static/object gate used the wrong variable | **Fixed** (`7460c7e`) |
| 2 | consumers missing `target_set_optimization_flags` | **Partially fixed** (`803ac19`) — see below |
| 3 | `CMAKE_BUILD_TYPE STREQUAL "Debug"` is a no-op under multi-config | **Not addressed** — deliberate; I accept the call |
| 4 | un-capped `maxRank = 10` | **Withdrawn** — I was wrong |
| 5 | `nMbptInputs` not capped | **Withdrawn** as a timeout risk |
| 6 | forced `SEQUANT_LTO=ON` on a no-LTO toolchain | **Not addressed** — deliberate; I accept the call |
| 7 | result variables under generic/reserved names | **Fixed** (`73ca17f`) — one sub-claim of mine was overstated |
| 8 | global MBPT context mutated in a benchmark body | **Fixed** (`e976a19`, `192a150`) |

**#1 verified empirically.** Repro mirroring `sequant_add_library` (STATIC) plus an executable, both through `target_set_optimization_flags`:

```
-- Performing Test SEQUANT_FAT_LTO_FLAG_SUPPORTED - Failed
-- REPRO: lib_copts=o1-NOTFOUND exe_copts=-flto=auto exe_lopts=-flto=auto
```

The static library now gets no LTO flags where fat objects are unavailable, the executable still gets them, and the whole thing builds and links. The bitcode-only-archive hazard is gone.

**#2 residual.** `python/CMakeLists.txt` and `tests/integration/eval/CMakeLists.txt` were covered. `doc/examples/CMakeLists.txt` (line 48, `target_set_warning_flags` with no optimization call) is still the one consumer without it — now only a consistency gap, since #1's fix means the archives are only LTO'd where a plain link still works. Inline note anchored to `tests/integration/eval/CMakeLists.txt` since that file has no hunks in this diff.

**#3 and #6 remain unaddressed by your deliberate choice, and I accept both.** #3 is real but low-impact and the complexity isn't obviously worth it. On #6, "explicit user choice, explicit consequences" is a fair policy; the only thing I'd still suggest is that the CMake error a user actually sees never mentions `SEQUANT_LTO`.

**Copilot's `batch`-lambda thread resolves in your favour.** `std::size_t const batch = 1;` is a const integral with a constant initializer, so reading it inside the lambda is not an odr-use and no capture is required. Copilot is wrong; the code is correct as written.

**#7, partial retraction of my own reasoning.** The three cache variables are properly prefixed now. My sub-claim about `CMAKE_SUPPORTS_COMPILER_LTO` was overstated: it is a non-cached function-local that `check_ipo_supported` overwrites on every call, so the parent-shadowing scenario I described cannot apply to it. Naming nit only.

Withdrawals for #4 and #5 are posted as replies on their original threads.

Four new findings on the new commits below, all low or medium-low. One of them (`python/CMakeLists.txt:16`) exists only because of my own bad advice in the previous review — pybind11 already applies LTO to that target, which I should have checked before asking for the call to be added. Sorry for the churn.

benchmarks/coupled_cluster.cpp
evaleev: **Withdrawn — I was wrong here.**

`git log --follow benchmarks/coupled_cluster.cpp` turns up `65e5c1b6 [skip ci] [benchmarks] trim the max rank of CC to 10`, which lowered `maxRank` from 20 to 10. So rank 10 was a deliberate trim from a working baseline, not an untested ceiling. @ajay-mk is right that it is fine in Release.

My "hours-to-weeks" number came from extrapolating the rank-4 timing in `operators.rst`, which measures something else entirely; that extrapolation was unfounded and I shouldn't have built a finding on it. Nothing here needs changing.

(The one narrow part that does hold is unrelated to the timing: at the merge base the benchmark threw because #456 removed the registry — which is exactly what `99e75f8` in this PR fixes.)

benchmarks/wick.cpp
evaleev: **Withdrawn as a timeout risk.** You're right that it causes no issue in practice, and I can now say why rather than just conceding.

The new CI step runs `--benchmark_dry_run=true`, i.e. a single iteration per benchmark, so the uncapped `nMbptInputs = 4` costs one `vac_av` of case 3 rather than a timed run. Both Debug jobs on `192a150` finished in about 15 minutes against the 60-minute budget, with the benchmark step included. The failure mode I described doesn't exist.

What's left is only a consistency nit, not worth a change on its own: the two `mbpt_vac_av` families are the only benchmarks in the tree that ignore `SEQUANT_QUICK_BENCHMARKS`, so "quick" isn't quite what it says. Fine to leave as is.

tests/integration/CMakeLists.txt
evaleev: Status: **partially fixed** by `803ac19`, plus one retraction.

**Retracting the `python-sequant` half.** I asked for `target_set_optimization_flags(python-sequant)` without checking whether pybind11 already handled it. It does — `pybind11_add_module` links `pybind11::lto` (`pybind11NewTools.cmake:324`) unless `NO_EXTRAS`, and that target carries `-flto[=thin|=auto]` plus `-fno-fat-lto-objects` on GCC as interface compile/link options, already guarded by a not-Debug genex. Following my advice put a second `-flto` spelling and a contradicting `-ffat-lto-objects` on the same command line. I've filed that as a new finding on `python/CMakeLists.txt:16` suggesting the line be reverted. My mistake, sorry for the churn.

**The `tests/integration/eval` half was right** and is now covered.

**Still outstanding:** `doc/examples/CMakeLists.txt:48`. With finding 1 now fixed this is no longer a breakage path — the archives are only LTO'd where fat objects exist, so those targets link fine without `-flto` — it's just the one remaining consumer built without the project's optimization settings.

cmake/compiler.cmake
evaleev: Fixed by `73ca17f` for the three cache variables — thanks.

One sub-claim of mine was overstated and I want to correct it rather than leave it standing: I lumped `CMAKE_SUPPORTS_COMPILER_LTO` in with the others, but it isn't a cache entry. `check_ipo_supported(RESULT ...)` writes a normal function-local variable that gets overwritten on every call, so the parent-shadowing scenario I described cannot apply to it. The reserved `CMAKE_` prefix is a naming nit, nothing more.

(That it is recomputed on every call is a separate matter — I've filed that as its own finding on `compiler.cmake:81`.)

cmake/compiler.cmake
Krzmbrzl: GNU `ld` accepts GNU-style LTO options for compatibility for pretty much exactly this reason. From the output of `ld --help`:
>   -flto                       Ignored for GCC LTO option compatibility

We'll keep as-is and can deal with any issues should they appear. These days, I don't think there are that many non-LTO capable linkers in-use anymore.

doc/user/getting_started/installing.rst
Krzmbrzl: Whether or not something is a cache variable is absolutely irrelevant to the docs. It's still a flag the user can specify and hence it is documented as such. It not being a cache variable is on purpose as this allows us to check whether the user set it and if not apply our default procedure.

tests/integration/eval/CMakeLists.txt
Krzmbrzl: Intentional. LTO can take nontrivial time during linking and there is no point in spending this time for optimized examples that are only built to check whether they compile & link without errors (aka:  they are never _really_ run)

evaleev (approved):

cmake/compiler.cmake
```diff
@@ -60,3 +60,130 @@ function(target_set_warning_flags TARGET)
...
+	# ar and ranlib index just fine - no special archiver needed
+	set(SEQUANT_LTO_AWARE_ARCHIVER TRUE)
+endif()
+
+function(target_set_optimization_flags TARGET)
+	if (NOT PROJECT_IS_TOP_LEVEL)
+		# When SeQuant is consumed as a subproject, how its targets are optimized is the
+		# encompassing project's call, not ours (cf. target_set_warning_flags)
+		return()
+	endif()
```
Krzmbrzl: @evaleev I disagree with this change. If `SEQUANT_LTO` is set, we should respect that. We can default to off in cases where it is not set and we are not the top-level project but when it is set, it shouldn't matter whether we are top-level or not. However, I think our defaults of only enabling with fat-object support is save and therefore I would in fact leave the defaults as-is and just remove the top-level check.

The big difference to warnings is that those don't influence the final binary and therefore they are only a development aid. Optimization flags on the other hand definitely affect the final product and therefore copying the warning function's behavior is inappropriate (imo).

evaleev: Agreed on the first half, and fixed in 49404d5 — you're right that this was inconsistent with itself: the early return sat *above* the `if (NOT SEQUANT_LTO STREQUAL "")` branch whose own comment says "always honor explicit user choice", so `-DSEQUANT_LTO=ON` in a subproject build did exactly nothing, silently. Explicit choice now wins regardless of scope; only the automatic per-target defaults remain top-level-only.

On removing the check entirely I'd still rather keep the default: I agree the fat-object gate makes the *linkability* safe either way, but the thing a parent project gets saddled with isn't just linkability, it's the build time and the link-time memory. That seems like the encompassing project's call to make when nobody asked for LTO explicitly, and now it can ask, which is the part that was actually broken.

One thing worth knowing if you still want the check gone: the archiver substitution was gated on the same `PROJECT_IS_TOP_LEVEL`, so dropping only the one in the function leaves `SEQUANT_LTO_AWARE_ARCHIVER` stuck at `FALSE` in subproject builds. I checked what that does on macOS — AppleClang leaves `CMAKE_CXX_COMPILER_AR` empty, so the `elseif (APPLE)` branch is the only one that sets the flag, and it was unreachable off the top level:

```
$ cmake -S outer -B build -DCMAKE_BUILD_TYPE=Release -DSEQUANT_LTO=ON   # SeQuant via add_subdirectory
CMake Error at compiler.cmake:149 (message):
  Requested LTO for static library 'mylib' but your compiler supports neither
  "fat" LTO objects nor an LTO-aware archiver (CMAKE_CXX_COMPILER_AR).
```

…even though cctools' `ar` indexes Apple bitcode fine. 49404d5 gates the archiver block on a new `SEQUANT_MAY_USE_LTO` instead, which keeps the two decisions in sync whichever way we settle the default. Matrix on AppleClang, before → after:

| scope | `SEQUANT_LTO` | before | after |
|---|---|---|---|
| top-level | *(auto)* | exe `-flto=auto`, lib none | unchanged |
| top-level | `ON` | both `-flto=auto` | unchanged |
| top-level | `OFF` | none | none (archiver no longer swapped) |
| subproject | *(auto)* | none | none |
| subproject | `ON` | **none** | both `-flto=auto`, archiver OK |
| subproject | `OFF` | none | none |

And yes — the warnings analogy doesn't carry, you're right about that; the argument for keeping a conservative default here is cost and ownership, not that it's harmless.

evaleev: @Krzmbrzl your eyes / signoff would be appreciated

Krzmbrzl: Will do on Monday or so

doc/user/getting_started/installing.rst
```diff
@@ -110,6 +110,12 @@ Useful CMake Variables
...
+     - empty (decide per target type): `ON` for static/object libraries if the compiler supports "fat" LTO objects, `OFF` otherwise. `ON` for
+       other target types (shared libs, executables, etc.)
+     - Controls whether SeQuant will be built with link-time optimizations (LTO) in non-debug builds (`CMAKE_BUILD_TYPE` != `Debug`). Set to
+       `ON` or `OFF` to override the per-target-type default for all SeQuant targets. Only has an effect when SeQuant is the top-level project;
+       when SeQuant is consumed as a subproject, optimization flags are left to the encompassing project.
```
Krzmbrzl:
```suggestion
       `ON` or `OFF` to override the per-target-type default for all SeQuant targets.
```

Krzmbrzl:
LGTM now

evaleev:
Pushed four small follow-ups on top of 32296c7, each independent:

- **c8ef3b1** `cmake/compiler.cmake`, three things:
  - `CMAKE_TRY_COMPILE_TARGET_TYPE` is now saved/restored around the three LTO probes instead of `unset()`. Hoisting the probes to directory scope meant the `unset` also discarded a value from a toolchain file or an encompassing project, after which every later `try_compile` in SeQuant (`vgkit_check_*`, Boost probes) tried to link an executable. Verified with a toolchain file: before `STATIC_LIBRARY → <undefined>`, after `STATIC_LIBRARY → STATIC_LIBRARY`.
  - The "neither fat objects nor an LTO-aware archiver" fatal error moved inside the manual-`-flto` branch. It doesn't apply to the `INTERPROCEDURAL_OPTIMIZATION` fallback (CMake substitutes the archiver itself there), but it fired on it anyway — MSVC + `SEQUANT_LTO=ON` aborted on the first static library with advice about llvm-ar.
  - Object libraries no longer get `-ffat-lto-objects`. Their objects only ever end up in a link step of ours (the `unit_tests-sequant-*-obj` into `unit_tests-sequant`, which carries `-flto` itself), so the machine-code half was never used, while producing it costs a second full compilation of every unit-test TU on GCC in every Release build. They're now treated like executables: LTO wherever `-flto` works, no fat objects.
- **f2187e2** `doc/examples/CMakeLists.txt`: examples get `target_set_optimization_flags` **only when `SEQUANT_LTO` is set explicitly**. Your "no point paying LTO link time for examples" default is kept as-is; but with an explicit `ON` on a toolchain without fat objects the SeQuant archives are bitcode-only, and the examples were the one consumer linked without `-flto` — reproduced on Linux/clang-12: `/usr/bin/ld: libSeQuant-symb.a: error adding symbols: file format not recognized`.
- **30532d8** `CMakeLists.txt`: `SeQuant-eval-ta` gets the warning and optimization flags — it's created outside `sequant_add_library` and was the only compiled library without either (Copilot's earlier point). `result.cpp` is clean under the full warning set with AppleClang; if GCC 14 disagrees, CI will say so.
- **dddf45d** `installing.rst`: restores the half of the dropped sentence that's still true (auto default is OFF as a subproject) alongside the new fact (explicit setting always applies).

Before/after flag matrix (static/object/exe × top-level/subproject × auto/ON/OFF) checked on AppleClang 17 and g++-15; only the object-library cells changed. Feel free to drop any of these if you disagree — they're separate commits for that reason.

doc/user/getting_started/installing.rst
```diff
@@ -110,6 +110,12 @@ Useful CMake Variables
      - ``ABORT`` in ``Debug`` mode, ``IGNORE`` otherwise
      - Controls how assertions within SeQuant's code are handled. Valid options are ``ABORT``, ``THROW`` and ``IGNORE``. The latter disables
        assertions, whereas the former keep them active and either abort the program or throw an exception on violation respectively.
+   * - SEQUANT_LTO
+     - empty (decide per target type): `ON` for static libraries if the compiler supports "fat" LTO objects, `OFF` otherwise. `ON` for
+       other target types (shared libs, executables, etc.). `OFF` for all targets when SeQuant is consumed as a subproject.
+     - Controls whether SeQuant will be built with link-time optimizations (LTO) in non-debug builds (`CMAKE_BUILD_TYPE` != `Debug`). Set to
+       `ON` or `OFF` to override the per-target-type default for all SeQuant targets; an explicit setting is honored regardless of whether
+       SeQuant is the top-level project.
```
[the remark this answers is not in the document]
evaleev: Right — and this corrects something I said earlier in this review: I had withdrawn the `python-sequant` half of my original finding on the grounds that pybind11 already applies LTO to the module. That's true only inside pybind11's own envelope, which is narrower than ours in two ways (pybind11 v3.1.0, `tools/pybind11Common.cmake:347` and `pybind11NewTools.cmake:228`):

- its "not debug" genex is `NOT (Debug OR RelWithDebInfo)`, whereas `target_set_optimization_flags` only skips `Debug`;
- `pybind11::lto` isn't linked at all when `CMAKE_INTERPROCEDURAL_OPTIMIZATION` is *defined*, with any value.

So `RelWithDebInfo` + `SEQUANT_LTO=ON` (or that variable defined + `SEQUANT_LTO=ON`) on a toolchain without fat objects leaves the module linked without `-flto` against bitcode-only archives. Both cases are off the CI matrix, which is why it's green. Verified with `RelWithDebInfo -DSEQUANT_LTO=ON`: module link flags before `-Xlinker -undefined -Xlinker dynamic_lookup`, after `-flto=auto -Xlinker -undefined -Xlinker dynamic_lookup`.

Fixed in 7522e3e the same way as the doc examples: `target_set_optimization_flags` on the module only `if (SEQUANT_LTO)`, so the automatic default still defers to pybind11. Where both apply the duplicate `-flto` is harmless — the module isn't archive-like, so there's no fat/no-fat clash.

cmake/compiler.cmake
```diff
@@ -60,3 +60,155 @@ function(target_set_warning_flags TARGET)
...
+# project may have set CMAKE_TRY_COMPILE_TARGET_TYPE for a reason (typically cross-compiling without
+# a usable link step), and since this runs at directory scope every later try_compile in SeQuant
+# would inherit whatever we leave behind.
+if (DEFINED CMAKE_TRY_COMPILE_TARGET_TYPE)
+	set(_sequant_saved_try_compile_target_type "${CMAKE_TRY_COMPILE_TARGET_TYPE}")
+endif()
+set(CMAKE_TRY_COMPILE_TARGET_TYPE "STATIC_LIBRARY")
+check_cxx_compiler_flag("-flto" SEQUANT_LTO_FLAG_SUPPORTED)
+check_cxx_compiler_flag("-flto=auto" SEQUANT_LTO_AUTO_SUPPORTED)
+check_cxx_compiler_flag("-flto;-ffat-lto-objects" SEQUANT_FAT_LTO_FLAG_SUPPORTED)
```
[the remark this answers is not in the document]
evaleev: This is the same point I raised earlier in the review (thread on line 96 of the previous revision), and it's technically accurate: with `CMAKE_TRY_COMPILE_TARGET_TYPE=STATIC_LIBRARY` none of the three `check_cxx_compiler_flag` probes link, so `SEQUANT_LTO_FLAG_SUPPORTED` says nothing about the linker. `SEQUANT_CAN_RELY_ON_CMAKE_LTO` (from `check_ipo_supported`) *does* link a static-lib + executable project and would catch a plugin-less linker — but it's OR'd in rather than AND'd, so it can never veto the manual path.

On the earlier rebuttal that `ld --help` says `-flto` is ignored: that's about `ld` being handed `-flto` directly. The failure mode is the driver (`clang++ -flto` / `g++ -flto`) asking `ld` to load `LLVMgold.so` / `liblto_plugin.so`, which is what fails when the binutils weren't built with plugin support. Exposure is low on current distros, so I'd call this a one-line `AND` rather than a blocker — but I'll leave that to @Krzmbrzl since it's his thread. I have not changed it.

PR #601, opened 2026-08-28
More UCC Features

ajay-mk:
## Changes
- Use Hamiltonian-matrix assembly for all UCC right-hand EOM equations, retaining per-block truncation ranks and scalar subtraction on diagonal blocks.
- Keep traditional CC on its existing connected `Hbar * R` path.
- Add `hbar_singles_comm_rank` to apply an additional singles-only similarity transform after the primary BCH expansion.
- Store the engine configuration as `CC::Options` and add `CC::with()` for constructing validated configuration variants.
- Reject orbital-optimized configurations that include singles and Bernoulli configurations that request the additional singles transform.
- Update the API documentation


SeQuant/domain/mbpt/models/cc.cpp
```diff
@@ -37,59 +37,69 @@ namespace sequant::mbpt {
...
-  if (hbar_expansion_ == HbarExpansion::Bernoulli && ansatz_ != Ansatz::U)
+CC::CC(size_t n, const Options& opts) : N(n), opts_(opts) {
+  if (opts_.hbar_expansion == HbarExpansion::Bernoulli &&
+      opts_.ansatz != Ansatz::U)
     throw Exception("CC: Bernoulli expansion requires the U ansatz");
-  if (unitary() && !hbar_comm_rank_)
+  if (opts_.hbar_expansion == HbarExpansion::Bernoulli &&
+      opts_.hbar_singles_comm_rank > 0)
+    throw Exception(
+        "CC: hbar_singles_comm_rank is not supported with Bernoulli");
```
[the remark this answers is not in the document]
ajay-mk: Fixed in [c537f7a](https://github.com/ValeevGroup/SeQuant/pull/601/commits/c537f7a9e9c627ae1b762ced50ffee21f9d0a03e)

evaleev (commented):
Review of the UCC EOM / `CC::Options` / Bernoulli changes. Six inline comments below, ordered by severity. The first two concern `hbar_singles_comm_rank` being applied in `hbar()` only, which makes the perturbed equations and the per-block EOM semantics inconsistent; the rest are test coverage and cleanup.

SeQuant/domain/mbpt/models/cc.cpp
```diff
@@ -37,59 +37,74 @@ namespace sequant::mbpt {
...
     return bernoulli::hbar(N, truncation, skip_singles());
 
   // for a non-unitary ansatz this is the cheaper connected-product form, which
   // is only equivalent to the commutator once the caller supplies operator
   // connectivity to ref_av (see lst_options() and the @warning on hbar())
-  return mbpt::lst(H(), T(N, skip_singles()), truncation, lst_options());
+  auto result = mbpt::lst(H(), T(N, skip_singles()), truncation, lst_options());
+
+  // Apply the optional singles-only transform after the primary BCH series.
+  if (opts_.hbar_singles_comm_rank > 0) {
```
evaleev, line 95: The σ1 wrap is applied here in `hbar()` only. `tʼ()` and `λʼ()` build `h1_bar` with a plain `lst(Hʼ, T(N), ...)`, and `rdm()` builds `lst(replacer, T(N), ...)` the same way, so with `hbar_singles_comm_rank > 0` Eq. 34's `h1_bar + [hbar, Tʼ]` mixes two different similarity transforms (e^{-σ1}e^{-σ}(·)e^{σ}e^{σ1} for H̄ vs e^{-σ}(·)e^{σ} for H̄ʼ), and `rdm()` silently ignores the option. Neither path throws.

Either propagate the wrap to every `lst` in the class (a small helper `similarity_transform(op, rank)` that does both steps would keep them in sync), or have `tʼ`/`λʼ`/`rdm` reject the option like the Bernoulli checks do.

SeQuant/domain/mbpt/models/cc.cpp
```diff
@@ -439,7 +455,7 @@ std::vector<ExprPtr> CC::eom_r_blocked(
   // Bernoulli H̄ is tensor-level, BCH H̄ operator-level; the bra/ket/vev trio
   // below must match it. Connectivity is empty either way, as everywhere on the
   // unitary path; only the operator-level vev forwards screen/use_topology.
-  const bool tensor_level = hbar_expansion_ == HbarExpansion::Bernoulli;
+  const bool tensor_level = opts_.hbar_expansion == HbarExpansion::Bernoulli;
 
   // One H̄ per distinct truncation order, reduced to its R part (Bernoulli only;
```
evaleev, line 463: Every per-block `hbar(k)` below also receives the σ1 wrap when `hbar_singles_comm_rank > 0`. The `eom_r` docs in `cc.hpp` promise that a rank-0 block is the bare Hamiltonian integrals, but with the option it becomes H + [H, σ1] (and the scalar subtracted on the diagonal is the vev of that). The `hbar()` docstring also never says the wrap is applied regardless of `truncation_rank`.

If this is intended, please document it in both places; either way a test combining the option with `block_ranks` would be good since nothing covers that path.

SeQuant/domain/mbpt/models/cc.cpp
```diff
@@ -411,26 +427,26 @@ container::svector<std::pair<std::int64_t, std::int64_t>> eom_manifolds(nₚ np,
...
   const auto K = manifolds.size();
   // `block_ranks` is read at i * K + j, so the ascending order above is what
-  // makes `{2,1,1,0}` mean SS, SD, DS, DD; empty means uniform hbar_comm_rank
+  // makes `{2,1,1,0}` mean SS, SD, DS, DD; empty means the configured H̄ rank,
+  // or the fourth commutator when no rank is configured.
   const std::vector<size_t> ranks =
-      block_ranks.empty() ? std::vector<size_t>(K * K, hbar_comm_rank().value())
-                          : block_ranks;
+      block_ranks.empty()
+          ? std::vector<size_t>(K * K, hbar_comm_rank().value_or(4))
```
evaleev: `hbar_comm_rank().value_or(4)` is unreachable here: `eom_r_ucc` is only entered for a unitary ansatz, and the constructor throws for a unitary ansatz without `hbar_comm_rank`. The comment above ("or the fourth commutator when no rank is configured") and the matching sentence in the private doc in `cc.hpp` describe a case that cannot occur. The previous `.value()` was the honest form.

SeQuant/domain/mbpt/models/cc.cpp
```diff
@@ -37,59 +37,74 @@ namespace sequant::mbpt {
...
   // is only equivalent to the commutator once the caller supplies operator
   // connectivity to ref_av (see lst_options() and the @warning on hbar())
-  return mbpt::lst(H(), T(N, skip_singles()), truncation, lst_options());
+  auto result = mbpt::lst(H(), T(N, skip_singles()), truncation, lst_options());
+
+  // Apply the optional singles-only transform after the primary BCH series.
+  if (opts_.hbar_singles_comm_rank > 0) {
+    auto opts = lst_options();
+    opts.skip_clone = true;
+    result = mbpt::lst(result, op::t(1), opts_.hbar_singles_comm_rank, opts);
```
evaleev: Nit: every other `lst` in this class spells the cluster operator via the `T(rank, skip_singles)` helper. `T(1, false)` here keeps the singles wrap in step with the primary expansion if that helper ever gains a wrapper.

tests/unit/test_mbpt_cc.cpp
```diff
@@ -268,9 +312,7 @@ TEST_CASE("mbpt_cc", "[mbpt/cc][valgrind_skip]") {
     REQUIRE(size(ip[0]) == 32);
     REQUIRE(size(ip[1]) == 11);
 
-    // one manifold means one block, whose <i|H̄|i> minus <0|H̄|0> is exactly the
-    // commutator the uniform path builds; the two assemble blocks by
-    // independent code, so this pins one against the other
+    // Explicit and default ranks agree for a single manifold too.
     const CC bch(2, {.ansatz = CC::Ansatz::U, .hbar_comm_rank = 2});
```
evaleev: Now that all UCC `eom_r` go through matrix assembly, this compares the same code path with itself: `eom_r(..., {2})` and `eom_r(...)` both run `eom_r_ucc`, so it no longer cross-checks the commutator form against the blocked form. The Bernoulli qUCCSD cases above pin term counts, but the BCH-UCC EOM output (whose semantics changed here, residual terms dropped) has no absolute check anywhere in this section; only uniform==blocked and mixed!=uniform. A `size()` check or `EquivalentTo` against an explicitly written expression for one small BCH block would catch e.g. a wrong sign on the diagonal scalar subtraction or a broken `i*K+j` mapping.

tests/unit/test_mbpt_cc.cpp
```diff
@@ -184,8 +229,7 @@ TEST_CASE("mbpt_cc", "[mbpt/cc][valgrind_skip]") {
     REQUIRE_THROWS_AS(CC(2, {.hbar_comm_rank = 2,
                              .hbar_expansion = CC::HbarExpansion::Bernoulli}),
                       Exception);
-    // only ranks 0..4 are implemented.
-    REQUIRE_THROWS_AS(bernoulli::hbar(2, 5, false), Exception);
+    REQUIRE_NOTHROW(bernoulli::hbar(1, 5, false));
```
evaleev: `bernoulli::hbar(1, 5, false)` builds a fifth-order Bernoulli H̄ (16 nested-commutator paths, up to five Wick commutators each) on every run, outside the `SEQUANT_SKIP_LONG_TESTS` guard, which runs against the PR's own "guard expensive UCC comparisons" goal. `nested_commutator_coefficients(5)` already exercises the recurrence at order 5 cheaply; if the no-throw check on `hbar` itself is wanted, it could move under the long-test guard.

SeQuant/domain/mbpt/models/cc.cpp
ajay-mk: The singles-wrap option is intended for unitary $\bar H$ only. t′ and rdm now reject it and λ′ is already restricted to traditional CC. 4104e56cd

SeQuant/domain/mbpt/models/cc.cpp
ajay-mk: Restored .value() and removed the fallback wording in 8a2de4206

SeQuant/domain/mbpt/models/cc.cpp
ajay-mk: Fixed in f1b20d986.

SeQuant/domain/mbpt/models/cc.cpp
ajay-mk: Documented the rank-0 behavior in 8a2de4206. Each block calls `hbar(k)`, where the singles wrap is already tested, so I did not add a separate test.

tests/unit/test_mbpt_cc.cpp
ajay-mk: Removed the self-comparison in 4e7409cab.

tests/unit/test_mbpt_cc.cpp
ajay-mk: Removed the full fifth-order `bernoulli::hbar` check in 3a3819042. The order-5 coefficient check remains.

ajay-mk:
This should be good to go now

evaleev (approved):

PR #605, opened 2026-09-03
Introduce Windows CI

Krzmbrzl:

PR #610, opened 2026-09-11
Add AGENTS.md: conventions for automated contributors

evaleev:
The repository has had no agent-facing conventions file, so every automated contributor rediscovers these from scratch — and in one case relearned the first one the hard way.

Split out of #596, where the `.expected` rule was agreed but would only have taken effect once that PR merged. Standalone so it applies to every branch now.

## What's here

**Never regenerate `*.expected` fixtures.** The suite can only check that today's output matches bytes a human previously blessed; it cannot check that a regenerated equation still *computes the right numbers*. So regenerating doesn't fix the failing comparison — it deletes the signal that a human needs to re-verify, and turns a red build into a false green. Stated unconditionally, with the rationalizations that don't justify it spelled out. Says where each tool writes its output so the diff can actually be inspected (ctest only prints "Files differ").

**Do not attribute tools in commits or PRs.** No `Co-Authored-By` trailer, no "Generated with Claude Code" line, no equivalent for any other assistant. Written down because several agents' harness defaults instruct the opposite, so without a project rule in the tree every session relitigates it — and the rule is stated as overriding those defaults rather than as a preference to weigh against them.

**Throw `sequant::Exception` or a class derived from it, nothing else.** Names the existing subclasses (`IndexSpace::bad_key`, `SerializationError`, `ConversionException`, `bad_any_comparable_cast`) so they are not flattened. Uniform in the tree since #612 merged (library, `utilities/`, `tests/`, `benchmarks/`), so this records the convention rather than proposing one. Also documents all three `SEQUANT_ASSERT_BEHAVIOR` values, their defaults, and why local builds should use `THROW`.

**Check includes with a non-unity build.** Some CI cells set `CMAKE_UNITY_BUILD=ON`, so a `.cpp` with a missing `#include` can compile there off a neighbouring TU and fail elsewhere. Points at `cmake.yml` for which cells, and gives the cheap local check (flip the option, rebuild one TU) rather than a from-scratch configure.

**Tests.** `check-sequant` / `ctest -R "^sequant"` is the full suite; `unit_tests-sequant` alone does not run the fixture comparisons, integration programs, or Python tests.

**Formatting** goes through `bin/admin/clang-format.sh`, which pins the version and knows the Homebrew paths, rather than a hard-coded `clang-format-17`.

**Layout**: a table of the module targets, their source directories, and their link dependencies taken from `target_link_libraries`, including `bliss`, the `eval::{ta,btas,tapp}` backend export names, and the `SeQuant/core/` directory vs `SeQuant-core` target distinction.

## `CLAUDE.md`

A regular file that imports `AGENTS.md` via `@AGENTS.md`, so Claude Code and tooling that reads `AGENTS.md` see one document. Not a symlink: a committed symlink checks out on Windows as a plain file containing the literal target path unless `core.symlinks` is set, and #605 is bringing Windows CI in.

## Notes for review

Docs only; no build or test impact.

Every factual claim was checked against the tree: the matrix in `cmake.yml`, the `SEQUANT_ASSERT_BEHAVIOR` block in `CMakeLists.txt`, the `add_test` entries under `tests/` and `utilities/`, `bin/admin/clang-format.sh`, and the `target_link_libraries` calls the layout table is drawn from.

Two things worth a maintainer's opinion:

- **Scope.** Kept to conventions that are already true and verifiable in-tree, rather than anything aspirational.
- **Teeth for the fixture rule.** It is prose only. A CI check that fails when a `*.expected` file changes without a maintainer label, or a `commit-msg` hook in `.pre-commit-config.yaml` rejecting attribution trailers, would be the mechanical versions if you want them.


AGENTS.md
```diff
@@ -0,0 +1,126 @@
...
+twice and reported a green suite both times. The bytes were in fact correct,
+which is why the episode is worth recording: matching bytes were not evidence of
+numerical correctness, and reporting a green suite concealed a question that was
+supposed to reach a maintainer.
+
+## Throw `sequant::Exception`, not `std` exceptions
+
+Use `sequant::Exception` (from `SeQuant/core/utility/exception.hpp`) rather than
+`std::runtime_error` / `std::logic_error` / `std::invalid_argument` when
+throwing, including in the standalone utilities under `utilities/`.
```
Krzmbrzl, line 77: Maybe be explicit that any non-listed kind of exception is also not permitted.

AGENTS.md
```diff
@@ -0,0 +1,126 @@
...
+a non-unity build rather than trusting CI:
+
+```
+cmake -S <src> -B <build> -G Ninja \
+  -DCMAKE_BUILD_TYPE=Debug \
+  -DSEQUANT_TESTS=ON -DBUILD_TESTING=ON \
+  -DSEQUANT_PYTHON=OFF -DSEQUANT_BUILD_DOCS=OFF -DSEQUANT_BENCHMARKS=OFF \
+  -DSEQUANT_COMPILE_DOC_EXAMPLES=OFF \
+  -DCMAKE_UNITY_BUILD=OFF
+ninja -C <build> unit_tests-sequant
```
Krzmbrzl, line 111:
```suggestion
cmake --build <build> --target unit_tests-sequant
```

AGENTS.md
```diff
@@ -0,0 +1,140 @@
...
+
+So whenever you add, move or prune includes, verify standalone compilation with
+a non-unity build rather than trusting CI:
+
+```
+cmake -S <src> -B <build> -G Ninja \
+  -DCMAKE_BUILD_TYPE=Debug \
+  -DSEQUANT_TESTS=ON -DBUILD_TESTING=ON \
+  -DSEQUANT_PYTHON=OFF -DSEQUANT_BUILD_DOCS=OFF -DSEQUANT_BENCHMARKS=OFF \
+  -DSEQUANT_COMPILE_DOC_EXAMPLES=OFF \
```
ajay-mk, line 109:
```suggestion
  -DSEQUANT_COMPILE_DOC_EXAMPLES=ON \
```

I think this should be turned on.

AGENTS.md
```diff
@@ -0,0 +1,140 @@
...
+
+Note the unit-test target is `unit_tests-sequant`, not `unit_tests`.
+
+## Formatting
+
+CI pins **clang-format 17** (`.github/workflows/formatting_check.yml`), and
+other major versions disagree with it. Check changed files with:
+
+```
+clang-format-17 --dry-run --Werror <files>
```
ajay-mk, line 122: Would it be better to directly use `clang-format.sh` here? In case at some point we bump clang-format version, just need to update one place.

AGENTS.md
```diff
@@ -0,0 +1,140 @@
...
+
+It exists so that callers can catch SeQuant-originated failures distinctly from
+other `std::exception`s, and it derives from `std::exception` (its ctor takes a
+`std::string`), so existing `catch (const std::exception&)` handlers still work.
+
+```cpp
+throw Exception("message");  // or sequant::Exception(...) if not in scope
+```
+
+The convention is uniform today — every throw in `SeQuant/` and `utilities/`
```
ajay-mk: There are few places where `std::runtime_error` is still used. Can be addressed in another PR. Issue https://github.com/ValeevGroup/SeQuant/issues/611

AGENTS.md
```diff
@@ -0,0 +1,140 @@
...
+uses it — so do not "consolidate" a file's throws toward `std::runtime_error`.
+Consolidate toward `Exception` instead.
+
+For internal invariants prefer `SEQUANT_ASSERT` over throwing. Note that
+`SEQUANT_ASSERT_BEHAVIOR` (`THROW` / `IGNORE`) changes whether assert-guarded
```
ajay-mk, line 92:
```suggestion
`SEQUANT_ASSERT_BEHAVIOR` (`THROW` / `IGNORE` / `ABORT`) changes whether assert-guarded
```

AGENTS.md
```diff
@@ -0,0 +1,140 @@
...
+paths execute, so a bug can be invisible under one setting and not the other.
+
+## A green CI build does not prove your includes are right
+
+CI builds most configurations with `CMAKE_UNITY_BUILD=ON`, which concatenates
```
ajay-mk, line 97:
```suggestion
CI builds debug configurations with `CMAKE_UNITY_BUILD=ON`, which concatenates
```

AGENTS.md
```diff
@@ -0,0 +1,140 @@
...
+
+- the diff is only renaming or reordering of intermediates;
+- the structure is unchanged (same contraction, alloc and intermediate counts);
+- the regenerated bytes turn out to match what someone else produced;
+- the change is "obviously" a cosmetic consequence of an intended fix.
+
+None of these establish that the emitted code still evaluates correctly, which
+is the only question the fixture exists to answer.
+
+This rule comes out of PR #596, where an agent regenerated five ITF fixtures
```
ajay-mk, line 53: I don't think we need to provide historical evidence in an instructions file. If agents look this up, we end up wasting tokens.

CLAUDE.md
```diff
@@ -0,0 +1,12 @@
+# CLAUDE.md
+
+This repository keeps its agent-facing conventions in `AGENTS.md`, which is the
+name most agent tooling looks for. Claude Code reads `CLAUDE.md`, so this file
+imports it — keep the content in `AGENTS.md`, not here.
+
+This is deliberately a regular file rather than a symlink to `AGENTS.md`: a
+committed symlink checks out on Windows as a regular file containing the
+literal text `AGENTS.md` unless `core.symlinks` is enabled, which silently
+yields an empty-looking config rather than an error. See #605.
```
ajay-mk, line 10:
```suggestion
yields an empty-looking config rather than an error.
```

AGENTS.md
Krzmbrzl: Could we instead instruct agents to just be aware that unity builds are used on CI and point them to the workflow file to figure it out themselves which exact ones use it? That seems more future proof.

ajay-mk:
Some suggestions from Claude's own `prompt-audit` skill (Using Opus 5)

[AGENTS.md.diff.txt](https://github.com/user-attachments/files/32122454/AGENTS.md.diff.txt)

AGENTS.md
evaleev: Done in 70d36501e: the section now reads "`sequant::Exception` or a class derived from it, nothing else", and names `std` types, string literals and unrelated classes as excluded.

AGENTS.md
evaleev: Switched to `cmake --build` for the check target. The one `ninja` line left is the Ninja-only `file.cpp^` single-TU rebuild, labelled as such.

AGENTS.md
evaleev: The from-scratch configure recipe is gone in 70d36501e. The section now says to flip `CMAKE_UNITY_BUILD=OFF` on the existing tree and rebuild the one TU, and to mirror the `BUILD_CONFIG` flags in `cmake.yml` (which set `SEQUANT_COMPILE_DOC_EXAMPLES=ON`) when configuring fresh, so the flag list is not duplicated here.

AGENTS.md
evaleev: Done: both the check and the fix command go through `bin/admin/clang-format.sh` now, and the version-suffixed binary name is gone.

AGENTS.md
evaleev: Done, with the defaults (`ABORT` for Debug, `IGNORE` otherwise), what CI uses, and a recommendation to configure local builds with `THROW`.

AGENTS.md
evaleev: Took the second suggestion: the text now says some CI configurations use unity builds and points at `.github/workflows/cmake.yml` for which, without enumerating them.

AGENTS.md
evaleev: Dropped the #596 paragraph.

CLAUDE.md
evaleev: `CLAUDE.md` is down to the import plus one sentence on why it is not a symlink; the paragraph this was on is gone.

evaleev:
Applied the substance of the prompt-audit diff in 70d36501e, with two departures: the Homebrew `llvm@17` path note is unnecessary now that the doc routes through `bin/admin/clang-format.sh` (which already prepends those paths), and the linear dependency chain is replaced by a table taken from `target_link_libraries`, since `export` links `optimize` and `mbpt` links `optimize` directly, which a chain cannot express.

AGENTS.md
evaleev: Those sites (`missing_instantiation_for`, `invalid_operand` / `unimplemented_method`, `bad_any_comparable_cast`, the string-literal throws in `mbpt`, and the `std::runtime_error` throws under `tests/`) are converted in #612. The text here no longer claims uniformity ahead of that; this PR should merge after #612.

AGENTS.md
```diff
@@ -0,0 +1,193 @@
...
+without it (see `.github/workflows/cmake.yml` for which is which). Whenever
+you add, move or prune includes, compile the affected TUs standalone before
+pushing.
+
+Unity is off by default in a local configure. If your build tree has it on,
+flip it and rebuild just the file:
+
+```
+cmake -DCMAKE_UNITY_BUILD=OFF <build>
+ninja -C <build> SeQuant/core/foo.cpp^      # Ninja: rebuild that one TU
```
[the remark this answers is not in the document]
evaleev: Confirmed on an out-of-source Ninja tree: neither the source-relative nor the build-dir-relative form is accepted (`ninja: error: unknown target`), because CMake writes absolute source paths into `build.ninja`. Only an absolute path works. f8de7ad0b documents `ninja -C <build> "$PWD/SeQuant/core/foo.cpp^"` from the source root and says why.

AGENTS.md
```diff
@@ -0,0 +1,193 @@
...
+per-session. If you have already committed with one, amend it out before
+pushing.
+
+## Throw `sequant::Exception`, nothing else
+
+Everything thrown in this tree — library, `utilities/`, `tests/`,
+`benchmarks/` — is `sequant::Exception` (`SeQuant/core/utility/exception.hpp`)
+or a class derived from it. Do not throw `std::runtime_error` / `std::logic_error` /
+`std::invalid_argument` or any other `std` type, string literals, or classes
+that do not derive from `Exception`. Helper functions that build an exception
```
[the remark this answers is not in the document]
evaleev: The branch was based on master from before #612, which converted all of those sites. Rebased onto current master in f8de7ad0b; the rule is true for the tree the PR now applies to.

AGENTS.md
```diff
@@ -0,0 +1,193 @@
...
+throw Exception("message");  // sequant::Exception outside namespace sequant
+```
+
+`Exception` derives from `std::exception` (its ctor takes a `std::string`), so
+`catch (const std::exception&)` handlers still work; it exists so that callers
+can catch SeQuant-originated failures distinctly. Where callers need to
+distinguish further, derive from it rather than reaching for a `std` type.
+`IndexSpace::bad_key`, `SerializationError`, `ConversionException` and
+`bad_any_comparable_cast` already do, and some code catches those by subtype —
+do not flatten them back into plain `Exception`.
```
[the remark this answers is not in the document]
evaleev: Same as above: `bad_any_comparable_cast` derives from `Exception` since #612 (tag.hpp), and the branch is now rebased onto that.

AGENTS.md
```diff
@@ -0,0 +1,193 @@
...
+back to docker:
+
+```
+bin/admin/clang-format.sh --dry-run --Werror <files>   # check
+bin/admin/clang-format.sh -i <files>                   # fix
+```
+
+`.pre-commit-config.yaml` runs the same wrapper and also rejects tabs, CRLF,
+U+00A0 and U+2013 in any file — but only once `pre-commit install` has been
+run in the clone.
```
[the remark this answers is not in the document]
evaleev: Reworded in f8de7ad0b: the tab check excludes `.out`, `.cmake`, `.js`, `.xml`, `.css`, and all the hooks run on the files being committed rather than on the whole tree.

evaleev:
On the suppressed Copilot note about Python tests: `test_export_python.cpp` is indeed part of `unit_tests-sequant`. The sentence meant the pybind11 binding tests registered by `python/CMakeLists.txt` (`sequant/unit/python/*`), which are separate ctest entries; f8de7ad0b says so explicitly.

PR #613, opened 2026-09-12
Batched array DAG evaluation (phase 1)

evaleev:
Phase 1 of batched array DAG evaluation for factorized CC equations.

The batched forest evaluator walks each root's binary tree and realizes batching by replaying a whole subtree once per slice of the batched index, with a node cache as the only link between roots and between slices. The DAG evaluator instead treats all roots as one DAG of distinct values, has the cost-model DP choose per value where in the nest of batch loops it is built (its home) so that a memory peak bound is met, compiles that into an explicit ordered schedule (a table of cells with loop nests, slice reads and scatter/sum assembles), and executes the table: every value is built once at its home level and consumers read slices of it, so cross-root and cross-slice sharing survives batching and the peak is set by loop placement rather than by replay.

What this adds (see `doc/dev/specs/2026-09-12-batched-array-dag-eval-as-built.md` for the as-built design):

- `optimize/cost_model.hpp`: batched DP over ordered cells with per-node external loop opens and bound-persistence charge; perf-first `(peak, flops, nsl)` frontier.
- `eval/peak_profile.hpp`, `lifetime_mask.hpp`, `legality.hpp`: loop identity (`compute_dag_boulevard`), per-occurrence facts, value identity (`value_key`), legality analysis.
- `eval/ordered_schedule.hpp`, `cell_table.hpp`, `cell_table_builder.hpp`: ordered schedule builder, explicit cell table with validator.
- `eval/ordered_executor.hpp`, `cell_registry.hpp`: table-driven executor (`compute_cell` reads operands through the table, `apply_one_op`, Sum/Scatter assembles, per-block scope guard for the backend's sparse-shape threshold).
- `eval/backends/dryrun/`: dry-run backend and metered walk/report.
- `eval/ordered_dump.hpp`: environment-gated diagnostics, documented in one place.
- Unit fixtures: dense, sparse and nested-tensor batched evaluation, dry-run walk/report, schedule and table gates.

Replaces #583 (same work, cleaned: whole-scope executor, cost-profile predictor, root-seed/cascade placement, `order_aware_recompute`/`node_level_placement`, placement router and remat removed; `CacheManager::store` -> `store_and_access`; `stamp_lifetime_masks` in `sequant::eval`). Kept for now: the batched forest descent (`BatchScheduler::forest_descent`).

Pins: TiledArray `1654b57b7` and BTAS `52aadfe54` (inherited from master through #616; the branch originally bumped TiledArray to `58acda635`, an ancestor). Companion mpqc4 PR: ValeevGroup/mpqc4#806 (merge order: this PR first, then mpqc4 repins to SeQuant master).

### Review-driven changes (after the initial squash)

- c75136d6b: a dry-run scatter block that neither appends nor prepends to the current coverage throws `sequant::Exception` regardless of the assertion policy; `LoopKey::color()` (12-bit packed slot, could collide) replaced by `std::hash<LoopKey>`; the ordered prepasses in `peak_profile.hpp`, `lifetime_mask.hpp`, `legality.hpp`, `value_node_map.hpp` walk with explicit stacks instead of recursion; value cells are bucketed by hash and confirmed structurally (`same_value`), with schedule-unique salted keys re-stamped onto the nodes, so a 64-bit collision cannot merge two values into one cell.
- f07a11696: the value-cell split propagated through the rest of the ordered path; `FullBinaryNode::size()` cached instead of recursive; the volatility walks in `meter.hpp` and `cache_manager.hpp` made iterative.
- f0b64c9e4: the value->node bridges hold forest nodes by pointer instead of copying subtrees (a 2000-node spine went from 13 GB RSS to 0.4 GB; real water-20 schedules -10%); move-assignment refreshes the source's cached size chain; range-yields-references `static_assert`s; the remaining four recursive walks in `eval.hpp` (incl. `subtree_any` on the ordered path) made iterative.
- b910a0ca7, 4990d8977: the test-side production-site check compares against `detail::collect_production_ids` (the notion `well_formed` uses: BuildSteps plus block outputs, so Reduction/LoopCarried values are not miscounted); self-move-assignment of a node is a no-op; three deep copies of the forest root list on the live path replaced by pointer lists; a test helper walks forest trees first-to-last.
- 7f253607d: dry-run roofline traffic = |left| + |right| + |result| (was left + a constant).
- b29c6e2f8, 8882d7e09: default member initializers on the optional `std::function`/`std::optional` fields that aggregate and designated initializers omit (`PeakModel`/`PeakBatchedModel`/`CostParams`/`OptimizeOptions::inner_pow`, `EvalStat::mem_left/right`, `Read::invariant_on`); GCC's `-Wmissing-field-initializers` reports these under master's `-Werror` while clang does not, verified by a g++ `-fsyntax-only` sweep of all 116 translation units.
- Master merged in twice rather than rebased (e13146fd1: #595 optimization flags, two test-file conflicts; e9009ae5a: #616 assert-policy seeding, pin conflict), so the commit shas cited in the review threads stay valid.

### Node identity vs evaluation order (CSE comparator; f1b83fbc1, 4b965902c)

Evaluation order stays exactly as the DP/binarizer emits it: cost, and peak memory in particular, is order-dependent, so the schedule has to be order-dependent and the tree is never permuted. Node identity, on the other hand, must not read that order. A Product node's children are therefore compared through a canonical view: `canonical_children(n)` returns the two children ordered by `canonical_operand_cmp`, which is O(1): a scalar operand sorts last (matching how `binarize` builds a `scalar * tensor` node), then ascending node id (the operand-order-independent `EvalExpr` hash), and a tie resolves to the emitted order. It reads only the two nodes' already-computed data. `TreeNodeEqualityComparator` compares `canonical_children(lhs)` against `canonical_children(rhs)` pairwise, one recursion per child, so the comparison is linear, unconditionally; nothing is reordered and nothing is cached on the node. This replaces the earlier unordered either-pairing match (exponential in product depth).

Resolving a node-id tie any other way was wrong: a fallback to a full equality comparison of the operand subtrees (as in f1b83fbc1) runs at every visit and makes the comparator T(n) = 4 T(n/2) = O(n^2) once hash-equal operands nest (a `t2 * t2` self-contraction reaches it); 4b965902c deletes it. Returning 0 is right for the case that occurs: two operands that are one value have one node id, either order is canonical for them, and an ordered comparison of two swapped spellings succeeds either way. The only other way to reach the tie is a 64-bit collision between distinct values, where the cost is a missed fold (two cache entries instead of one), never a wrong one, since equality still compares both subtrees in full.

The same reasoning applies to the value key the batched DAG scheduler derives from a node and its operands' keys: it combined the operand keys in emitted order and so did not fold swapped spellings even where the node cache did. Combining them in ascending order for Product nodes removes 10 duplicate builds from the water-20 aux+occ DAG schedule (1542 -> 1532 cold, 1286 -> 1276 warm). Evidence that the fold is right, since fewer builds alone proves nothing: `[w20-auxocc-walk]`, the strict dry-run walk over that schedule with lobound, range and fill-once checks in an assert-live build; a water-20 (aux-c/occ-e/occ-c batched CSV-CCSD) wet run is the end-to-end check: at a TA sparse-shape threshold of 1e-10, occ batch 8 and occ batch 16 converge to -5.010110275693 and -5.010110275767, i.e. batch-size independent to 7e-11 (at 1e-8 they differ by 6e-8 and sit 3e-6 above, so the screening threshold, not the batching, is what moves the energy). An unbatched water-20 reference does not fit the box, so losslessness is judged by batch-size independence; occ batch 8 peaks at 33 GB against 61 GB for batch 16. The water-8 and he10 wet gates fold nothing and are regression gates only. The generated NEVPT2 ITF is byte-identical: the export path keys CSE off `TreeNodeHasher` + `TreeNodeEqualityComparator` and never computes a DAG value key, and the canonical view folds a subset of what the unordered match folded, so export CSE lost nothing.


evaleev:
@Krzmbrzl would you please numerically verify if regenerated nevpt2 exports are correct? These are the only failing cases at the moment: https://github.com/ValeevGroup/SeQuant/actions/runs/34776721253/job/103776060655?pr=613#step:14:741

Krzmbrzl:
Why have they changed again? I thought we sorted the symmetry thing out 👀

But yeah, will do

evaleev:
For the nevpt2 verification: this is not the symmetry change again. The only mechanism behind the fixture diff is the CSE comparator: `TreeNodeEqualityComparator` (`SeQuant/core/eval/eval_node_compare.hpp`, ~line 136, from c3b2c07a1) now treats a `Product` node's two operands as commutative, so `(DF:Faa * Ym1:aa) * DF:Faa`, which the binarizer emits in one operand order inside R1 and the opposite order in the CSE shared by R2u/R2, hashes and compares equal. Previously it was two values (`CSE25:aa[vw]` for the R2 pair, an inline `I:aa[ux]` in R1); now it is one `CSE7:aa[ux]` with three uses, hoisted to the earliest consumer, and every later CSE renumbers.

Net effect on the emitted ITF: one fewer alloc, contraction and load, two fewer drops, same 65 stores and 49 CSE statements, no term added or removed, no coefficient changed; 22 of the 24 diff hunks are pure renumbering. Disabling only that commutative branch reproduces the old output byte for byte. The diff hunks with the mapping are in the PR branch's ctest output (`utilities/external-interface/examples/nevpt2_v2_cse_merged.itfaa` next to the `.expected`).

SeQuant/core/eval/eval_node_compare.hpp
```diff
@@ -51,77 +56,121 @@ struct TreeNodeEqualityComparator {
...
+      if (lhs->op_type() && *lhs->op_type() == EvalOp::Product) {
+        bool const in_order = (*this)(lhs.left(), rhs.left()) &&
+                              (*this)(lhs.right(), rhs.right());
+        bool const swapped = (*this)(lhs.left(), rhs.right()) &&
+                             (*this)(lhs.right(), rhs.left());
+        if (!in_order && !swapped) {
+          return false;
+        }
+        return true;
       }
```
Krzmbrzl: To me it seems that doing things this way make the calls _really_ expensive. This now scales exponentially for every factor in a product.

The better fix would be to ensure that the assumption of the previously existing implementation actually holds. That is, equations must not be factored as AB in one place and BA in another.

Presumably, this requires tweaking with how we do comparisons. Using hashes makes for a nifty optimization but also makes it hard to impossible to get any specific order. A change in the comparison philosophy could also address #614. My proposal would be to do deep comparison when doing operator<=> and only use the hash-based shortcut for equality comparison. Given that the latter is the most frequent anyway, this should retain most of the performance boost whilst giving us a predictable and controllable sort order.

Krzmbrzl (changes requested):

SeQuant/core/eval/eval_node_compare.hpp
evaleev: You're right, and the first version of the fix was still wrong on exactly the point you raised, so thank you for pushing on it. Two commits, f1b83fbc1 and 4b965902c.

Evaluation order stays exactly as the DP/binarizer emits it (cost, and peak in particular, is order-dependent, so the schedule has to be); what was wrong is that node identity was reading that order. A Product node's children are now compared through a canonical view (`canonical_children`, ordered by an O(1) key: scalar last, then ascending node id, ties keep emitted order), and the comparator compares the two views pairwise, one recursion per child, so it is linear unconditionally; nothing is reordered or cached on the node. The value key the DAG scheduler derives gets the same treatment. Details, the O(n^2) trap the first version had, and the evidence for the extra folds are in the PR description under "Node identity vs evaluation order".

Could you take another look?

SeQuant/core/eval/backends/dryrun/result.hpp
```diff
@@ -0,0 +1,881 @@
...
+    } else {
+      auto& lohi = it->second;
+      bool const append = block_lo == lohi.second;
+      bool const prepend = block_hi == lohi.first;
+      SEQUANT_ASSERT(append || prepend);
+      if (append)
+        lohi.second = block_hi;
+      else
+        lohi.first = block_lo;
+    }
```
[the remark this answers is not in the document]
evaleev: Fixed in c75136d6b: a scatter block that neither appends nor prepends to the current coverage (gap or overlap) now throws sequant::Exception, independent of the assertion policy.

SeQuant/core/eval/dag_scope.hpp
```diff
@@ -0,0 +1,107 @@
...
+
+  /// \brief A single opaque color encoding the FULL loop identity (\c depth AND
+  /// \c loop_slot), for use where one \c std::size_t must distinguish loops:
+  /// \c ordered_n_batches_by_loop (one batch count PER LOOP, not per
+  /// loop-group -- two members of one group are DISTINCT loops with distinct
+  /// batch counts; see ordered_executor.hpp). Keying on \c depth alone would
+  /// conflate same-group sibling loops. (\c loop_slot < 4096 in every
+  /// realized schedule.)
+  [[nodiscard]] std::size_t color() const {
+    return (depth << 12) | static_cast<std::size_t>(loop_slot);
```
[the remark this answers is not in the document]
evaleev: Fixed in c75136d6b: color() is gone; ordered_n_batches_by_loop() and the other loop-keyed tables key on LoopKey itself through std::hash<LoopKey> (full-width fields, no packing), so slot numbering is unbounded without collisions.

SeQuant/core/eval/peak_profile.hpp
```diff
@@ -0,0 +1,1223 @@
...
+    }
+
+    if (detail::dump_enabled("SEQUANT_DUMP_OPENS") &&
+        !n->batch_loops_opened_here().empty())
+      detail::dump_opens(n);
+
+    container::svector<std::size_t> child_recs;
+    if (!n.leaf()) {
+      child_recs.push_back(self(self, n.left(), child_ectx, child_opener));
+      child_recs.push_back(self(self, n.right(), child_ectx, child_opener));
```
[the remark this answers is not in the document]
evaleev: Fixed across c75136d6b (this prepass plus the analogous walks in lifetime_mask.hpp, legality.hpp, value_node_map.hpp), f07a11696 (meter.hpp, cache_manager.hpp, the volatility walks; FullBinaryNode::size() now cached instead of recursive) and f0b64c9e4 (the four remaining recursive walks in eval.hpp, including subtree_any on the ordered path). All are explicit-stack traversals now; a hidden N=20000 spine test in test_peak_profile.cpp exercises the ordered prepasses, and the value->node bridges hold forest nodes by pointer (f0b64c9e4), which took the same spine from 13 GB RSS to 0.4 GB.

SeQuant/core/eval/peak_profile.hpp
```diff
@@ -0,0 +1,1223 @@
...
+      for (std::size_t pre : r.ectx_opener)
+        o.ectx_opener_point.push_back(pre_to_point.at(pre));
+      o.contracted_batched = r.contracted_batched;
+      o.opens = r.opens;
+      o.operand_points = r.operand_points;
+      o.loop_slot = r.loop_slot;
+      o.reduced_slot = r.reduced_slot;
+      return o;
+    };
+    auto const it = hash_to_cell.find(r.key);
```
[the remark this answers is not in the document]
evaleev: Fixed in c75136d6b (propagated through the rest of the ordered path in f07a11696): value grouping now buckets by hash and confirms membership structurally (same_value), so two distinct values with colliding hashes get distinct cells; the cells carry schedule-unique salted keys that are re-stamped onto the nodes, and b910a0ca7 adds a test-side check that every non-leaf cell has exactly the production sites the schedule's own well-formedness rules require.

evaleev:
Merged master (4112eb6c0) into this branch as e13146fd1 rather than rebasing, so the commit shas cited in the review threads stay valid. Two test-file conflicts resolved; master's -Werror on the unit tests (#595) needed a default member initializer on PeakModel/PeakBatchedModel::inner_pow plus three test-only cleanups (details in the merge commit message). RelWithDebInfo build clean, unit suite green (318143 assertions / 296 cases).

SeQuant/core/eval/eval_node_compare.hpp
```diff
@@ -23,11 +30,95 @@ struct TreeNodeHasher {
...
+/// subtrees in full and rejects a mismatch. Resolving such a collision instead
+/// would mean comparing the operand subtrees here, and a full comparison per
+/// visit turns the comparator quadratic on nested hash-equal operands (a
+/// `t2 * t2` self-contraction is enough): T(n) = 4 T(n/2). Keeping this O(1)
+/// is what makes the comparator provably linear.
+///
+/// \return <0 if \p a belongs first, >0 if \p b does, 0 if the two are
+///         interchangeable (read them in the order they were emitted).
+template <typename TreeNode>
+int canonical_operand_cmp(TreeNode const &a, TreeNode const &b) {
```
Krzmbrzl: This should be `operator<=>` returning one of the canonical comparison types (e.g. `std::strong_ordering`)

SeQuant/core/eval/eval_node_compare.hpp
```diff
@@ -51,77 +142,135 @@ struct TreeNodeEqualityComparator {
...
+      }
 
-    if (lhs->type_id() != rhs->type_id()) {
-      return false;
-    }
+      // Cheapest discriminator first: hash::value is the EvalExpr's cached
+      // node id (a load), size() the subtree's cached node count (also a load
+      // now, see FullBinaryNode::size_ -- it used to WALK the subtree, which
+      // made every probe of a structurally-keyed map over a deep tree
+      // quadratic in tree size). Both are necessary conditions, so the order
```
Krzmbrzl: You should review all your comments and remove explanations of what used to be the case in favor of only stating what is. Explanations on the used-to state should be part of the commit message but it doesn't help anyone making sense of current code that isn't familiar with the old code (and hence is unlikely to be helpful as an in-code comment)

SeQuant/core/eval/eval_node_compare.hpp
```diff
@@ -51,77 +142,135 @@ struct TreeNodeEqualityComparator {
...
+      // The recursive child comparison is still REQUIRED and is NOT redundant
+      // with the graph check above: the connectivity graph encodes only the
+      // immediate two factors' connectivity, not each factor's recursive
+      // build, so two products with the same immediate graph but different
+      // sub-values must still be told apart. Product operand subtrees are
+      // bounded in depth (a contraction of a fixed set of factors), so they
+      // stay recursive.
+      if (lhs->op_type() && *lhs->op_type() == EvalOp::Product) {
+        auto const [lfirst, lsecond] = canonical_children(lhs);
+        auto const [rfirst, rsecond] = canonical_children(rhs);
```
Krzmbrzl:
```suggestion
        const auto &[lfirst, lsecond] = canonical_children(lhs);
        const auto &[rfirst, rsecond] = canonical_children(rhs);
```

evaleev: I don't think this does anything since canonical_children returns a pair of refs: https://godbolt.org/z/e9esvafhd

Krzmbrzl: It doesn't but it makes it more obvious that these variables are in fact references

SeQuant/core/eval/eval_node_compare.hpp
evaleev: Done in 635f0f6af: canonical_operand_cmp returns std::strong_ordering (scalar operand greater, then by node id, equal on a tie), and canonical_children orders on it.

SeQuant/core/eval/eval_node_compare.hpp
evaleev: Applied in 635f0f6af.

SeQuant/core/eval/eval_node_compare.hpp
evaleev: Agreed. Done for this file in 635f0f6af and, since the same habit was all over the branch, for every file the branch touched in 393455c45, cad640767 and e9183bda6: comments now state what the code does and why, with no references to earlier arrangements, removed components, commit shas or review rounds; the history stays in the commit messages. The same pass also aligned the Doxygen style with each file's existing convention and dropped emphasis by capitals.

PR #615, opened 2026-09-14
Constant: order by value, not by hash

### LINKED ISSUE #614: `Constant` comparison fragile
We have the test
https://github.com/ValeevGroup/SeQuant/blob/442867b5b61d6fe731cf40099df9430aa39c649b/tests/unit/test_utilities.cpp#L416-L419
that aligns with user's expectation. However, when looking into the details, we find that `Constant::operator<` is also using the hash-based comparison. Using the hash instead of the value itself for comparison gives us no guarantee that constants compare as we would expect. And indeed, on Windows using MSVC, this test breaks (see #605).

evaleev:
Closes #614.

`Constant` did not override `Expr::static_less_than`, so `Constant::operator<` fell through to the base implementation, which compares hash values. Hash order is unrelated to numeric order and differs between platforms; the `ExprMatcher` test that expects `Constant(1) < Constant(2)` held on macOS and Linux by accident and broke under MSVC (#605).

This orders `Constant`s by value instead: by real part, then by imaginary part (both exact rationals).

**No fixture impact:** `Sum` canonicalization sorts summands by hash and only consults `operator<` on exact hash ties, and `Product` folds constants into its scalar, so no canonical or generated output changes.

**Test:** new block in the `constant` section of `tests/unit/test_expr.cpp` checks strict ordering over negative, fractional, integer and complex constants, irreflexivity, and the imaginary-part tie-break. It failed on macOS before the fix (the bug was latent there too) and passes after. Full `unit_tests-sequant` passes.

Related, not addressed here: `Variable` also lacks a `static_less_than` override and orders by hash.


SeQuant/core/expressions/constant.cpp
```diff
@@ -75,4 +75,11 @@ bool Constant::static_equal(const Expr &that) const {
   return value() == static_cast<const Constant &>(that).value();
 }
 
+bool Constant::static_less_than(const Expr &that) const {
+  const auto &that_value = static_cast<const Constant &>(that).value_;
```
Krzmbrzl, line 79: Is this unconditional cast safe? Aka: is the input expression always guaranteed to be a constant?

Krzmbrzl (approved):

SeQuant/core/expressions/constant.cpp
evaleev: yes

